PR #52, opened 2024-07-04
Add health check and development script

### LINKED ISSUE #49: Health endpoint
For people deploying behind a load balancer, a simple healthcheck endpoint that verifies connection to the database would be very helpful

will2hew:
Adds a simple health check endpoint that verifies the connection to the database. 

This also contains a development server script I personally find helpful, but am happy to remove if you'd prefer. 

Closes #49

### DIFF
--- a/apps/server/src/integrations/health/health.controller.ts
+++ b/apps/server/src/integrations/health/health.controller.ts
@@ -1,0 +1,26 @@
+import { KyselyDB } from '@docmost/db/types/kysely.types';
+import {
+  Controller,
+  Get,
+  HttpCode,
+  HttpStatus,
+  InternalServerErrorException,
+  Post,
+} from '@nestjs/common';
+import { sql } from 'kysely';
+import { InjectKysely } from 'nestjs-kysely';
+
+@Controller()
+export class HealthController {
+  constructor(@InjectKysely() private readonly db: KyselyDB) {}
+
+  @Get('health')
+  @HttpCode(HttpStatus.OK)
+  async health() {
+    try {
+      await sql`SELECT 1=1`.execute(this.db);
+    } catch (error) {
+      throw new InternalServerErrorException('Health check failed');
+    }
+  }
+}
--- a/apps/server/src/integrations/health/health.controller.ts
+++ b/apps/server/src/integrations/health/health.controller.ts
@@ -5,22 +5,33 @@ import {
   HttpCode,
   HttpStatus,
   InternalServerErrorException,
-  Post,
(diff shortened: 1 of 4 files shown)


apps/server/src/integrations/health/health.controller.ts
Philipinho, line 1: Looks good. We need Redis check too. 
Redis is linked with queues, collaboration and socket.io.

You can use the logger to log the health check on error.

package.json
```diff
@@ -12,7 +12,8 @@
     "client:dev": "nx run client:dev",
     "server:dev": "nx run server:start:dev",
     "server:start": "nx run server:start:prod",
-    "email:dev": "nx run @docmost/transactional:dev"
+    "email:dev": "nx run @docmost/transactional:dev",
+    "dev": "pnpm concurrently -n \"frontend,backend\" -c \"cyan,green\" \"pnpm run client:dev\" \"pnpm run server:dev\""
```
Philipinho, line 16: LGTM.

apps/server/src/integrations/health/health.controller.ts
will2hew: Sounds good, done

will2hew:
@Philipinho I have added the requested redis check and logging. 

I would note, that currently the entire application crashes when Postgres and/or Redis connections are lost, so this health check does perform its job - we should probably make the app fail in a more graceful way. I think that should be a follow up to this however.

apps/server/src/integrations/health/health.controller.ts
```diff
@@ -5,22 +5,33 @@ import {
...
   @HttpCode(HttpStatus.OK)
   async health() {
     try {
+      const redis = new Redis(this.environmentService.getRedisUrl());
+
       await sql`SELECT 1=1`.execute(this.db);
+      await redis.ping();
     } catch (error) {
-      throw new InternalServerErrorException('Health check failed');
+      this.logger.error('Health check failed');
```
Philipinho, line 33: One last thing. We should add the `error` object to the log so we can know what went wrong.

Philipinho:
> that currently the entire application crashes when Postgres and/or Redis connections are lost
It only crashes when the database connection fails during startup after 10 retries. Once the app is running, lost db connections do not crash it.

A lost Redis connection will not crash it, both during startup and while running. Redis retries indefinitely.

Philipinho:
I could not add my commits here. I created a new branch #64. 
I will merge the new branch after you review it (If you've got time).

PR #56, opened 2024-07-05
Remove explicit credentials configuration for S3 client

### LINKED ISSUE #50: Utilise the AWS client's native credential parsing
We currently provide the s3-client with credentials. The AWS NodeJS SDK will get these automatically if we don't provide it to the client. 

The downside of the current approach, is we can't use best-practice short lived AWS (or other) S3 credentials, which contains the extra `AWS_SESSION_TOKEN` environment variable.

This should simply be fixed by deleting this code, and the SDK will get the credentials automatically.

will2hew:
By explicitly providing the credentials configuration block, we were overwriting the AWS Node SDK's built in environment variable parsing. This meant that when using short lived tokens, which uses `AWS_SESSION_TOKEN` as well as `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` that it was failing, as it would see the access key and the access key ID, but not a valid session. 

By removing it, we let the Node SDK natively get the credentials. This works with both the legacy (and not recommended) long-term credentials, providing both the access key and ID, but also with the more secure recommended credentials that utilise a session token. 

The docs will need to be updated to reference `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` rather than `AWS_S3_SECRET_ACCESS_KEY` and `AWS_S3_SECRET_ACCESS_KEY`

Closes #50

### DIFF
--- a/apps/server/src/integrations/environment/environment.service.ts
+++ b/apps/server/src/integrations/environment/environment.service.ts
@@ -43,14 +43,6 @@ export class EnvironmentService {
     return this.configService.get<string>('STORAGE_DRIVER', 'local');
   }
 
-  getAwsS3AccessKeyId(): string {
-    return this.configService.get<string>('AWS_S3_ACCESS_KEY_ID');
-  }
-
-  getAwsS3SecretAccessKey(): string {
-    return this.configService.get<string>('AWS_S3_SECRET_ACCESS_KEY');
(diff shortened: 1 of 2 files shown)


Philipinho:
#71 closes this.

PR #57, opened 2024-07-05
CORS setup and Vite devServer proxy

will2hew:
We can use the devServer proxy option in Vite to avoid needing a CORS exception for development


Philipinho:
Proxy is a better choice. Thank you.

PR #58, opened 2024-07-05
Add request typings and change JWT extraction

will2hew:
This PR does a small bit of housekeeping to make it easier to potentially add OIDC in the future, and generally improves developer experience. 

- Adds a new `AppRequest` type which contains the correct typing for `user`, `workspace` and `workspaceId` and uses that new type where appropriate
- Uses the provided extractor functions option from the JWT Module


Justman100:
@sidnelui-krystal @wuast94 @Philipinho

What here now?

PR #64, opened 2024-07-05
Add health check

Philipinho:
Built upon PR #52 by @will2hew

This PR adds the `/api/health` endpoint.

I refactored the existing PR to make use of the more advanced [nestjs/terminus](https://www.npmjs.com/package/@nestjs/terminus) package.

### DIFF
--- a/apps/server/src/integrations/health/postgres.health.ts
+++ b/apps/server/src/integrations/health/postgres.health.ts
@@ -1,0 +1,38 @@
+import { InjectKysely } from 'nestjs-kysely';
+import {
+  HealthCheckError,
+  HealthIndicator,
+  HealthIndicatorResult,
+} from '@nestjs/terminus';
+import { Injectable, Logger } from '@nestjs/common';
+import { sql } from 'kysely';
+import { KyselyDB } from '@docmost/db/types/kysely.types';
+
+@Injectable()
+export class PostgresHealthIndicator extends HealthIndicator {
+  private readonly logger = new Logger(PostgresHealthIndicator.name);
+
+  constructor(@InjectKysely() private readonly db: KyselyDB) {
+    super();
+  }
+
+  async pingCheck(key: string): Promise<HealthIndicatorResult> {
+    let isHealthy = false;
+
+    try {
+      await sql`SELECT 1=1`.execute(this.db);
+      isHealthy = true;
+    } catch (e) {
+      this.logger.error(JSON.stringify(e));
+    }
+
+    if (isHealthy) {
+      return this.getStatus(key, isHealthy);
+    } else {
+      throw new HealthCheckError(
+        `${key} is not available`,
+        this.getStatus(key, isHealthy),
+      );
+    }
+  }
+}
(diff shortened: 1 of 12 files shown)


apps/server/src/integrations/health/redis.health.ts
```diff
@@ -0,0 +1,41 @@
...
+@Injectable()
+export class RedisHealthIndicator extends HealthIndicator {
+  private readonly logger = new Logger(RedisHealthIndicator.name);
+
+  constructor(private environmentService: EnvironmentService) {
+    super();
+  }
+
+  async pingCheck(key: string): Promise<HealthIndicatorResult> {
+    let isHealthy = false;
```
will2hew: `isHealthy` is unnecessary - it would be more conventional to attempt to return inside the try / catch block, or throw an error, for example:

```
    try {
      const redis = new Redis(this.environmentService.getRedisUrl(), {
        maxRetriesPerRequest: 15,
      });

      await redis.ping();
      return this.getStatus(key, true);
    } catch (e) {
      this.logger.error(e);
      throw new HealthCheckError(
         `${key} is not available`,
          this.getStatus(key, false),
      );
    }
```

apps/server/src/main.ts
```diff
@@ -38,7 +38,8 @@ async function bootstrap() {
     .addHook('preHandler', function (req, reply, done) {
```
will2hew, line 38: This confuses me a bit - is this not the job of the `DomainMiddleware` 

If its that you cannot throw NotFoundException inside the middleware, I'd instead suggest a globally applied `WorkspaceGuard` which can.

apps/server/src/integrations/health/postgres.health.ts
```diff
@@ -0,0 +1,38 @@
...
+@Injectable()
+export class PostgresHealthIndicator extends HealthIndicator {
+  private readonly logger = new Logger(PostgresHealthIndicator.name);
+
+  constructor(@InjectKysely() private readonly db: KyselyDB) {
+    super();
+  }
+
+  async pingCheck(key: string): Promise<HealthIndicatorResult> {
+    let isHealthy = false;
```
will2hew: Ditto comment of not needing the seperate `isHealthy` as I said on the redis health check

will2hew (changes requested):

apps/server/src/integrations/health/redis.health.ts
Philipinho: I modelled it after the terminus inbuilt indicators but didn't implement the timeout error instance which warrants the variables. Will clean it up.

apps/server/src/main.ts
Philipinho: It was more of the exception and the chain of request cycle.  I needed to put it at the top level before the guards. We could revisit this if you have a better idea that works. 
I know it's not clean enough but I struggled with Fastify's handling of middleware which led to this hack.

will2hew: I suspected that may be it - fastify inside NestJS is not quite as mature as express, we could consider returning to express but I'm unsure on why fastify was chosen. Either way unrelated to this PR :)

will2hew (approved):

will2hew:
LGTM

PR #69, opened 2024-07-07
Add SES mail driver

will2hew:


will2hew:
Opened accidentally; will look to upstream just the SES transport, not the entirety of my fork

Philipinho:
Thank you for contribution.
Tbh, I am not looking to add other email drivers outside of the existing SMTP and Postmark drivers.
SMTP should work fine as the universal driver.

PR #81, opened 2024-07-12
add STMP_SECURE and changed auth config

olivierIllogika:
Only configure auth object in smtp config if user and password are provided. If no auth object is defined, [nodemailer assumes the connection is already authenticated](https://nodemailer.com/smtp/#authentication).

Also exposed the secure config option.

### DIFF
--- a/apps/server/src/integrations/environment/environment.service.ts
+++ b/apps/server/src/integrations/environment/environment.service.ts
@@ -87,17 +87,6 @@ export class EnvironmentService {
     return parseInt(this.configService.get<string>('SMTP_PORT'));
   }
 
-  getSmtpAuth(): object {
-    let auth = undefined;
-    if (this.getSmtpUsername() && this.getSmtpPassword()) {
-      auth = {
-          user: this.getSmtpUsername(),
-          pass: this.getSmtpPassword(),
-        };
-    }
-    return auth;
-  }
-
   getSmtpSecure(): boolean {
     const secure = this.configService
       .get<string>('SMTP_SECURE', 'false')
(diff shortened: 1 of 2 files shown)


apps/server/src/integrations/environment/environment.service.ts
```diff
@@ -87,6 +87,24 @@ export class EnvironmentService {
...
+  getSmtpAuth(): object {
+    let auth = undefined;
+    if (this.getSmtpUsername() && this.getSmtpPassword()) {
+      auth = {
+          user: this.getSmtpUsername(),
+          pass: this.getSmtpPassword(),
+        };
+    }
+    return auth;
+  }
```
Philipinho: Thank you for your contribution.

Can you move the logic to the mail provider? https://github.com/docmost/docmost/blob/main/apps/server/src/integrations/mail/providers/mail.provider.ts

olivierIllogika: Like so?
```diff

     switch (driver) {
       case MailOption.SMTP:
+        let auth = undefined;
+        if (environmentService.getSmtpUsername() && environmentService.getSmtpPassword()) {
+          auth = {
+              user: environmentService.getSmtpUsername(),
+              pass: environmentService.getSmtpPassword(),
+            };
+        }
         return {
           driver,
           config: {
             host: environmentService.getSmtpHost(),
             port: environmentService.getSmtpPort(),
             connectionTimeout: 30 * 1000, // 30 seconds
-            auth: environmentService.getSmtpAuth(),
+            auth,
             secure: environmentService.getSmtpSecure(),
           } as SMTPTransport.Options,
         };

```

Philipinho: Yep. Thank you.

Philipinho:

PR #85, opened 2024-07-15
feat: implement Markdown and HTML page imports

Philipinho:
This PR makes it possible to import Markdown and HTML pages.
You can import multiple pages in a single import.

In a future release, there will be support for platform-specific imports (Confluence, Notion etc).
This is good enough for now.

<div align="center">
Page import menu item
</div>

---
<div align="center">
Page import modal
</div>


danz0l99:
I second that, would be much better to have a markdown importer or at least a way to do raw markdown that we can cut and paste articles into to get the same.

PR #243, opened 2024-09-03
Support I18n

lleohao:
## Description
Support for i18n has been added, and currently, support for Chinese has been implemented.

**Untranslated parts**

- Form error prompts
- Editor

## Related Issue
#18 

## How Has This Been Tested?
Modifications are limited to the display content; I tested the relevant pages.

## Translation file division

Divided according to the directories of features, for example, all files under `features/space` are placed in `space.json`.

The rest (common components, layout, navigation, etc.) are all placed in `translation.json`.

## Screenshots (if appropriate):
See below


lleohao:
## Login Page

Xnip2024-09-03_13-23-14

## Workspace Setting

Xnip2024-09-03_13-24-45
Xnip2024-09-03_13-24-54
Xnip2024-09-03_13-25-01
Xnip2024-09-03_13-25-08
Xnip2024-09-03_13-25-17
Xnip2024-09-03_13-25-23
Xnip2024-09-03_13-25-32
Xnip2024-09-03_13-25-54

lleohao:
## Profile Setting 

Xnip2024-09-03_13-26-18

## Preference Setting

Xnip2024-09-03_13-26-26

lleohao:
## Workspace Details

Xnip2024-09-03_13-26-45
Xnip2024-09-03_13-27-00
Xnip2024-09-03_13-27-07
Xnip2024-09-03_13-27-24
Xnip2024-09-03_13-27-29
Xnip2024-09-03_13-31-42

Philipinho:
Looks really cool.
I see you separated the translations by workspace, group, space etc. which makes it manageable. What are your thoughts on having single translation files per language?
I ask in cases of overlapping translations. Say we have a "Delete" string in both workspace and group translations, then we have to manually translate twice. Is my thinking correct?

lleohao:
> Looks really cool. I see you separated the translations by workspace, group, space etc. which makes it manageable. What are your thoughts on having single translation files per language? I ask in cases of overlapping translations. Say we have a "Delete" string in both workspace and group translations, then we have to manually translate twice. Is my thinking correct?

Using multiple translation files is a habit I developed from my work experience, as my project is a massive one, it naturally involves multiple translation files.

However, it seems that using multiple translation files is premature for `docmost` at the moment, as the benefits it brings are not significant. I can merge them all into one.

lleohao:
>  I ask in cases of overlapping translations. Say we have a "Delete" string in both workspace and group translations, then we have to manually translate twice. Is my thinking correct?

Yes, you need to translate "Delete" twice.

For complex projects, there may be cases where the same copy needs to be translated into different copies in different contexts, but it seems that `docmost` has not encountered such issues so far.

Philipinho:
Thank you for the insight.
We should go for a single file then.

lleohao:
@Philipinho I have consolidated the translation files into a single file.

Philipinho:
LGTM.

Philipinho:
After further testing, here are a few observations.

i. I see it loads all the translation files by default. Is it possible to only load the currently used translation file?
ii. We should use the full language code format. e.g. en-US, zh-CN etc
iii. We have the `locale` column in the user table. We should update it when the user changes their language, for persistence.

Please let me know if you need any help from my side. I appreciate your work.

lleohao:
> i. I see it loads all the translation files by default. Is it possible to only load the currently used translation file?

Since English is the default fallback language, it will still request the English translation file when you select Chinese, but it will not request the Chinese translation file if you select English.

lleohao:
@Philipinho save language preferences code at https://github.com/docmost/docmost/pull/243/commits/49900050f7901bcb0dcfa6c515a5463d186850f8

lleohao:
It seems I updated the wrong field, I'll come and fix it.

lleohao:
> @Philipinho save language preferences code at [4990005](https://github.com/docmost/docmost/commit/49900050f7901bcb0dcfa6c515a5463d186850f8)


It seems I updated the wrong field, I'll come and fix it.

apps/server/src/core/user/user.service.ts
Philipinho, line 1: There is already the `locale` column in the users table which should hold the language code. We should use that instead of the settings column preference.

lleohao: https://github.com/lleohao/docmost/commit/0488441fbbf87ed2ccd1cff2df7380c41e72a1ef

lleohao: @Philipinho I fixed it

Philipinho: Good job. Thank you.

Philipinho: Great job. Thank you.

danielalves96 (approved):

iamxiaojianzheng:
Could consider completing the translation work based on https://crowdin.com/ instead of manually maintaining the files?

ja49619:
> Could consider completing the translation work based on https://crowdin.com/ instead of manually maintaining the files?

There is also a popular platform webplate

Matinal111:
Looking forward to the update🫡

PR #250, opened 2024-09-03
feat(backend): forgot password

### LINKED ISSUE #182: Be able to reset user password
can you point it out how to implement this ?

aldy505:
Partially closes #182, still need to implement the frontend side of things.

How this works:
```http
POST http://API_URL/auth/forgot-password
Content-Type: application/json
{
  "email": "foo@example.com"
}
```

Then check user's email, a code should be provided. If their email don't exist, it will always return 200 OK response. This is intended to protect the privacy & security of the user.

Then, re-hit the endpoint.
```http
POST http://API_URL/auth/forgot-password
Content-Type: application/json

{
  "email": "foo@example.com",
  "code": "XXXXX",
  "newPassword": "very strong password"
} 
```

A security notice email will be sent if the code is correct. If not, a 400 BAD REQUEST response of "Incorrect code" will be sent.

### DIFF
--- a/apps/server/src/core/auth/services/auth.service.ts
+++ b/apps/server/src/core/auth/services/auth.service.ts
@@ -15,6 +15,10 @@ import { comparePasswordHash, hashPassword } from '../../../common/helpers';
 import { ChangePasswordDto } from '../dto/change-password.dto';
 import { MailService } from '../../../integrations/mail/mail.service';
 import ChangePasswordEmail from '@docmost/transactional/emails/change-password-email';
+import { ForgotPasswordDto } from '../dto/forgot-password.dto';
+import { TemporaryCodeRepo } from '@docmost/db/repos/temporary-code/temporary-code.repo';
+import { randomBytes } from 'crypto';
+import ForgotPasswordEmail from '@docmost/transactional/emails/forgot-password-email';
 
 @Injectable()
 export class AuthService {
@@ -22,6 +26,7 @@ export class AuthService {
     private signupService: SignupService,
     private tokenService: TokenService,
     private userRepo: UserRepo,
+    private temporaryCodeRepo: TemporaryCodeRepo,
     private mailService: MailService,
   ) {}
 
@@ -44,6 +49,95 @@ export class AuthService {
 
     const tokens: TokensDto = await this.tokenService.generateTokens(user);
     return { tokens };
+  }
+
+  async forgotPassword(
+    forgotPasswordDto: ForgotPasswordDto,
+    workspaceId: string,
+  ) {
+    const user = await this.userRepo.findByEmail(
+      forgotPasswordDto.email,
+      workspaceId,
+      true,
+    );
+    if (!user) {
+      return;
+    }
+
+    if (
+      forgotPasswordDto.code == null ||
+      forgotPasswordDto.newPassword == null
+    ) {
+      // Generate 5-character temporary code
+      const code = randomBytes(8).toString('hex').slice(0, 5).toUpperCase();
+      const hashedCode = await hashPassword(code);
+      await this.temporaryCodeRepo.insertTemporaryCode({
+        code: hashedCode,
+        user_id: user.id,
+        workspace_id: user.workspaceId,
+        expires_at: new Date(new Date().getTime() + 5 * 60 * 1000), // should expires in 5 minute
+      });
+
+      const emailTemplate = ForgotPasswordEmail({
+        username: user.name,
+        code: code,
+      });
+      await this.mailService.sendToQueue({
+        to: user.email,
+        subject: 'Reset your password',
+        template: emailTemplate,
+      });
+
+      return;
+    }
+
+    // Get all temporary codes that are not expired
+    const temporaryCodes = await this.temporaryCodeRepo.findByUserId(
+      user.id,
+      user.workspaceId,
+    );
+    // Limit to the last 3 codes, so we have a total time window of 15 minutes
+    const validTemporaryCodes = temporaryCodes
+      .filter((code) => code.expires_at > new Date() && code.used_at == null)
+      .slice(0, 3);
+
+    for (const code of validTemporaryCodes) {
+      const validated = await comparePasswordHash(
+        forgotPasswordDto.code,
+        code.code,
+      );
+      if (validated) {
+        // Update the used_at field to the current time
+        await this.temporaryCodeRepo.updateTemporaryCode(
+          {
+            used_at: new Date(),
+          },
+          code.id,
+        );
+
+        const newPasswordHash = await hashPassword(
+          forgotPasswordDto.newPassword,
+        );
+        await this.userRepo.updateUser(
+          {
+            password: newPasswordHash,
+          },
+          user.id,
+          workspaceId,
+        );
+
+        const emailTemplate = ChangePasswordEmail({ username: user.name });
+        await this.mailService.sendToQueue({
+          to: user.email,
+          subject: 'Your password has been changed',
+          template: emailTemplate,
+        });
+
+        return;
+      }
+    }
+
+    throw new BadRequestException('Incorrect code');
   }
 
   async register(createUserDto: CreateUserDto, workspaceId: string) {
--- a/apps/server/src/database/migrations/20240903T124647-temporary-code.ts
+++ b/apps/server/src/database/migrations/20240903T124647-temporary-code.ts
@@ -1,0 +1,26 @@
+import { sql, Kysely } from 'kysely';
+
+export async function up(db: Kysely<any>): Promise<void> {
+  await db.schema
+    .createTable('temporary_codes')
+    .addColumn('id', 'uuid', (col) =>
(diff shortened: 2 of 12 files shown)


apps/server/src/database/migrations/20240903T124647-temporary-code.ts
Philipinho, line 1: I suggest we rename the table to `user_tokens`.

Rename `code` to `token`
`expires_at` should be nullable.
We should add a new `type` varchar column to hold the token type.  The token type could be for password reset, email change etc

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +51,95 @@ export class AuthService {
...
+    if (!user) {
+      return;
+    }
+
+    if (
+      forgotPasswordDto.code == null ||
+      forgotPasswordDto.newPassword == null
+    ) {
+      // Generate 5-character temporary code
+      const code = randomBytes(8).toString('hex').slice(0, 5).toUpperCase();
```
Philipinho: We should generate the token with nanoIdGen() from https://github.com/docmost/docmost/blob/be0d97661ae83c115446cec48cdabd0ebb5efc40/apps/server/src/common/helpers/nanoid.utils.ts#L5

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +51,95 @@ export class AuthService {
...
+      .slice(0, 3);
+
+    for (const code of validTemporaryCodes) {
+      const validated = await comparePasswordHash(
+        forgotPasswordDto.code,
+        code.code,
+      );
+      if (validated) {
+        // Update the used_at field to the current time
+        await this.temporaryCodeRepo.updateTemporaryCode(
```
Philipinho: We should delete generated tokens on successful validation, and unused tokens within the same window.

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +51,95 @@ export class AuthService {
...
+      forgotPasswordDto.newPassword == null
+    ) {
+      // Generate 5-character temporary code
+      const code = randomBytes(8).toString('hex').slice(0, 5).toUpperCase();
+      const hashedCode = await hashPassword(code);
+      await this.temporaryCodeRepo.insertTemporaryCode({
+        code: hashedCode,
+        user_id: user.id,
+        workspace_id: user.workspaceId,
+        expires_at: new Date(new Date().getTime() + 5 * 60 * 1000), // should expires in 5 minute
```
Philipinho: Let's make it 1 hour.

Philipinho (changes requested):
I left some feedback. Thank you for your contribution.

apps/server/src/core/auth/services/auth.service.ts
aldy505: Isn't it too long? On average companies, this is usually less than 5 minutes.

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +51,95 @@ export class AuthService {
...
+  async forgotPassword(
+    forgotPasswordDto: ForgotPasswordDto,
+    workspaceId: string,
+  ) {
+    const user = await this.userRepo.findByEmail(
+      forgotPasswordDto.email,
+      workspaceId,
+      true,
+    );
+    if (!user) {
```
sahariardev, line 62: how about throwing exception if user not found

if (!user) {
      throw new NotFoundException('User not found');
    }

aldy505: That way you open up a security risk of knowing that "this user with this email address actually exists". On most security regulations (basically ISO27001 certification and stuff), providing that kind of data is not allowed.

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +51,95 @@ export class AuthService {
...
+        // Update the used_at field to the current time
+        await this.temporaryCodeRepo.updateTemporaryCode(
+          {
+            used_at: new Date(),
+          },
+          code.id,
+        );
+
+        const newPasswordHash = await hashPassword(
+          forgotPasswordDto.newPassword,
```
sahariardev, line 117: need to validate newPassword length?

aldy505: I wanted to create a password requirements. But since the current "change password" isn't guarded with anything (minimum characters, combination of alphanumerics, etc), I decided to go with the current "change password" implementation.

Philipinho: The validation is done in the dto `@MinLength(8)`. 
https://github.com/docmost/docmost/blob/be0d97661ae83c115446cec48cdabd0ebb5efc40/apps/server/src/core/auth/dto/change-password.dto.ts#L3-L7

aldy505: @Philipinho do we want to only have minimum characters length?

apps/server/src/core/auth/services/auth.service.ts
Philipinho: If an intruder has got access to the user’s email, 5 minutes or 1 hour will not make any difference.
For example, GitHub’s password reset is valid for 3 hours.

apps/server/src/core/auth/services/auth.service.ts
Philipinho: @aldy505 yes, minimum length validation is enough.

apps/server/src/database/repos/user-tokens/user-tokens.repo.ts
```diff
@@ -0,0 +1,89 @@
...
+    workspaceId: string,
+    tokenType: string,
+    trx?: KyselyTransaction,
+  ) {
+    const db = dbOrTx(this.db, trx);
+    return db
+      .selectFrom('userTokens')
+      .select([
+        'id',
+        'code',
```
Philipinho: We renamed `code` to `token`. It should reflect in its usage across too.

apps/server/src/database/types/db.d.ts
```diff
@@ -185,6 +190,17 @@ export interface Workspaces {
   updatedAt: Generated<Timestamp>;
 }
 
+export interface UserTokens {
+  id: Generated<string>;
+  code: string;
```
Philipinho: It should be renamed to `token`.
You can regenerate the db types with `pnpm nx run server:migration:codegen`.

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +50,94 @@ export class AuthService {
...
+      return;
+    }
+
+    if (
+      forgotPasswordDto.code == null ||
+      forgotPasswordDto.newPassword == null
+    ) {
+      // Generate 5-character user token
+      const code = nanoIdGen().slice(0, 5).toUpperCase();
+      const hashedCode = await hashPassword(code);
```
Philipinho: I think hashing the token is redundant. Since it passes through a mail server, the server admin will still have access to the raw token irrespective of hashing, which defeats the purpose of hashing.

Philipinho (commented):
Thank you for your work. We are almost there.

apps/server/src/core/auth/services/auth.service.ts
aldy505: I beg to disagree with that.
1. Keeping it hashed with bcrypt slows down any brute force or dictionary attacks if a malicious actor is doing the "forgot password" action without ever looking at the email.
2. If the email is sent via TLS/STARTLS, at least the content is encrypted in-transit. Whether it can be read by the server administrator is a different issue.
3. If the email is compromised, they have a bigger problem in the first place.

Philipinho: i. Rate limits, when implemented, should be more appropriate to defend against brute-force attacks.
ii. We are not exposing the token to other users. The tokens are ephemeral. Hashing does not solve any problem, it just increases complexity with little to no benefits.
It does not have the same effects as hashing user passwords.

apps/server/src/core/auth/services/auth.service.ts
```diff
@@ -46,6 +50,94 @@ export class AuthService {
...
+    if (!user) {
+      return;
+    }
+
+    if (
+      forgotPasswordDto.code == null ||
+      forgotPasswordDto.newPassword == null
+    ) {
+      // Generate 5-character user token
+      const code = nanoIdGen().slice(0, 5).toUpperCase();
```
Philipinho: We can do this with `nanoIdGen(5)`.

apps/server/src/core/auth/services/auth.service.ts
aldy505: Still, it's better to put more guards while we can rather than just passing the problem to some other place (in this case, an upstream reverse proxy that has rate limiter functionality). I don't think hashing the token is a bad thing to do.

Philipinho (approved):

PR #299, opened 2024-09-11
feat: Implement nodemailer ignore tls property

### LINKED ISSUE #211: [Bug / Feat Request] SMTP ignore SSL errors
Hi,

I have an internal smtp server that is used with mismatched hostnames on the SSL certificate. Is there an environmental variable available to turn off any SSL checks for SMTP?

Thank you.

Shadowfita:
Resolves #211 .


RiekertKBW:
Would love to have that soon.
Can't invite others to it because of this Error.

Philipinho:
Is there a way to achieve this without yet another env variable?

Shadowfita:
> Is there a way to achieve this without yet another env variable?

It's possible. I can't really think of a simpler way though, especially since there are other nodemailer props configured in env already.

PR #301, opened 2024-09-11
Issue fixed wrong page title

### LINKED ISSUE #223: [Bug] Title being saved for wrong page
1. Edit page title
2. Immediately switch to another page

Expected behaviour: first page title changed
Actual: new title is applied to second page

vivaldi_MIpPj1coof

jigarjoshi011:
fixes #223 
Solves issue while Edit page title Immediately switch to another page.
**Note: if the page change is less than half a second the page title will be not saved.**


Philipinho:
Should be better handled in the next release.

PR #325, opened 2024-09-18
feat: trash for deleted pages in space

### LINKED ISSUE #59: Feature Request - Recycle bin
Hi, love your work so far. Please keep on going.

It would really be helpful, if you were able to have a recycle bin for pages. Or maybe it can be added to the page history of the workspace / parent page.

mobley-trent:
- [x] Move to trash
- [x] Permanently delete
- [x] Restore
- [x] Only space admins can permanently delete
- [x] Cron to delete trashed pages older than 30 days.

Closes #59


mobley-trent:
POC


https://github.com/user-attachments/assets/e5668ef1-5b44-408b-86a3-43b09bb84d27

mobley-trent:
> I can’t seem to find how pages with children are handled. Ideally, when a page is soft-deleted, its children should be too. Same applies to restoration.
> 
> Let me know if I am missing something.

Deletion and restoration works with pages with subages as well.

mobley-trent:
https://github.com/user-attachments/assets/89b847e5-a0c6-4400-b72d-0015aa9914c9

Philipinho:
Aside from removing it from the tree, do sub-pages get soft-deleted too? i.e `deleted_at` field updated for each sub-page.

mobley-trent:
> Aside from removing it from the tree, do sub-pages get soft-deleted too? i.e `deleted_at` field updated for each sub-page.

I assumed the previous route for page deletion would also delete the subpages. I simply updated the route to update the `deletedAt` field.

Philipinho:
The default deletion process auto deletes child pages via foreign key relationship.

mobley-trent:
> The default deletion process auto deletes child pages via foreign key relationship.

Updated the `remove` (soft delete) and `restore` routes to fetch page ids for descendants as well.

mobley-trent:
Hi @Philipinho I would appreciate your feedback

Philipinho:
@mobley-trent thank you for your work. I will get to it.

mobley-trent:
Any update on this @Philipinho ?

mobley-trent:
PTAL @Philipinho

mobley-trent:
@Philipinho I opted to auto-reload the page to handle the state

Philipinho:
One of the issues I encountered the last time I tried it was, If a page with children gets deleted and the child of a deleted page gets restored, the child and it's children gets lost and does not appear in the sidebar.

apps/client/src/features/space/components/recycled-pages.tsx
```diff
@@ -22,10 +22,12 @@ export default function RecycledPagesList({
 
   const handleRestorePage = async (pageId: string) => {
     await restorePageMutation.mutateAsync(pageId);
+    window.location.reload();
```
Philipinho: I don't think this provides a great UI experience.

mobley-trent:
> One of the issues I encountered the last time I tried it was, If a page with children gets deleted and the child of a deleted page gets restored, the child and it's children gets lost and does not appear in the sidebar.

I handled this by setting the `parentPageId` to null when restoring a page. This effectively ensures that the restored page and child pages are restored as root pages in the sidebar.

apps/client/src/features/space/components/recycled-pages.tsx
mobley-trent: I was trying to find a way to hide the deleted nodes in the SpaceTree and unhide when you restore them, can we use this as a workaround ?

Philipinho: I lean towards mutating the tree on the client side. I will have a go at it when I work on the PR again.

Issue #257, opened 2024-09-04
[BUG] Drawio preview wrong svg text


kolanski:
## Bug Report: Cyrillic Font Issue in Diagram Preview

**Description:**

When creating diagrams using Cyrillic fonts in the Draw.io application, an issue occurs where the text appears corrupted or misencoded in the preview mode. However, the text displays correctly when the diagram is opened in the editor. This issue only affects the preview and does not impact the text in the editor view.

**Steps to Reproduce:**

1. Create a new diagram in Draw.io.
2. Use Cyrillic fonts or text within the diagram.
3. Save the diagram and switch to preview mode.

**Expected Result:**

The text in Cyrillic fonts should appear correctly and consistently in both the editor and preview modes.

**Actual Result:**

The text appears corrupted or misencoded in the preview mode, while it displays correctly in the editor mode.

**Screenshots:**

- Diagram with Cyrillic fonts in edit mode:
  ![png](https://github.com/user-attachments/assets/110baef1-1fee-4d63-8922-63a0ba67fd4a)
  
- Screenshot of the corrupted preview:
  ![diagram drawio2](https://github.com/user-attachments/assets/e57ba421-0da3-4a16-9eb4-7b03c6ff216d)

**Additional Information:**

- Issue occurs only in the preview mode.
- The editor view displays the text correctly.

**Severity:**

Moderate — affects the usability of the preview feature for diagrams containing Cyrillic text.

**Suggested Fix:**

Investigate the encoding or rendering process used in the preview mode. Maybe base64 svg import as blob converting is wrong.

xseth03:
Hi,

Same issue with latin font :
![image](https://github.com/user-attachments/assets/de635068-929b-4cc5-ad62-3d440e59ab02)

Servostar:
The issue happens due to an encoding error. Drawio makes use of UTF-8, while the client decodes to Latin-1, messing up non-ASCII characters.

Philipinho:
Fixed. Thanks to @Servostar.

PR #369, opened 2024-10-03
fix: base64 encoded drawio image decoded to Latin-1 instead of UTF-8

Servostar:
Drawio encodes the image data as Base64 for sending. The client decodes the Base64 to text assuming Latin-1 encoding, which is incompatible to UTF-8 used by Drawio messing up non-ASCII characters (see issue #257).

This fix wraps the decoding function `btoa` by decoding to binary first and encoding the binary to UTF-8.

This fix allows to use UTF-8 characters in drawio diagrams without messing up the output.

Solution based on: https://developer.mozilla.org/en-US/docs/Glossary/Base64

### DIFF
--- a/apps/client/src/lib/utils.ts
+++ b/apps/client/src/lib/utils.ts
@@ -53,11 +53,21 @@ export async function svgStringToFile(
   return new File([blob], fileName, { type: "image/svg+xml" });
 }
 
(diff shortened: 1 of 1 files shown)


aayushchugh (approved):
LGTM 🚀

PR #453, opened 2024-10-30
Fix sidebar responsivity

ftibi93:
This pull request fixes the responsivity of the sidebar. (found on 0.4.1, and I could reproduce the issue on the latest main)

# Current behavior

## Big screen size
Screenshot 2024-10-30 162324
Screenshot 2024-10-30 162513

The tree items don't fill up the full height of the sidebar.

## Small screen size

Screenshot 2024-10-30 162834
Screenshot 2024-10-30 163512

The items overflow so there are items which are not visible (the document "Last page" should be at the bottom of the list but it is below the viewport).

## Cause of the problem

The sidebar consists of 3 sections: workspace selector, menu, tree list. In the current implementation each section takes up as much space as they need but inside the tree list section there is a fix height set.

The `treeContainer` class has a fix height of `68vh` which causes the elements to either flow over the viewport which renders them inaccessible or to not fill up the available space in the sidebar.

## The solution

I left the first two sections unchanged and modified the tree list section to fill up the remaining space. Then changed the tree's height to be equal to its parent.

Screenshot 2024-10-30 163214
Screenshot 2024-10-30 163302

## Testing

- tested the fix with multiple browsers (Chrome, Firefox), multiple screen sizes and with one Android device (Firefox)
- made sure lazy loading still works
- on page refresh the tree scrolls to the opened document's position

P.S.
- this is my first attempt to contribute to an open source project, so I am not entirely sure that I did everything right
- this is my first time working with React/Nestjs (I am a Angular/Spring Boot developer)
- (and sorry for my English)

### DIFF
--- a/apps/client/src/features/space/components/sidebar/space-sidebar.tsx
+++ b/apps/client/src/features/space/components/sidebar/space-sidebar.tsx
@@ -134,7 +134,7 @@ export function SpaceSidebar() {
           </div>
         </div>
 
-        <div className={classes.section}>
+        <div className={[classes.section, classes.sectionPages].join(' ')}>
           <Group className={classes.pagesHeader} justify="space-between">
             <Text size="xs" fw={500} c="dimmed">
               Pages
--- a/apps/client/src/features/page/tree/components/space-tree.tsx
+++ b/apps/client/src/features/page/tree/components/space-tree.tsx
@@ -187,6 +187,11 @@ export default function SpaceTree({ spaceId, readOnly }: SpaceTreeProps) {
     }
   }, [treeApiRef.current]);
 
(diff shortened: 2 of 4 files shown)


apps/client/src/features/space/components/sidebar/space-sidebar.tsx
```diff
@@ -134,7 +134,7 @@ export function SpaceSidebar() {
           </div>
         </div>
 
-        <div className={classes.section}>
+        <div className={[classes.section, classes.sectionPages].join(' ')}>
```
Philipinho, line 137: Let's use `clsx` here for consistency.

PR #533, opened 2024-12-04
feat: adding resend provider to deal easily with emails

iagocavalcante:
In this PR the idea is insert a new provider to send emails, the Resend.com provider


sibelius (approved):

Philipinho:

I am not looking to add new mail drivers, as the SMTP driver should be enough in most cases. However, you can maintain your own fork if integrating Resend directly is best for your usecase.

Issue #570, opened 2024-12-19
Clicking on emojis from different pages will display multiple emoji lists, one for each selected page


omarNaifer12:
https://github.com/user-attachments/assets/de6e2dba-0e5a-4699-88fb-8ddce3cf2271

omarNaifer12:
Hi @Philipinho , can I work on this?

Philipinho:
Hi @omarNaifer12, I am happy to review your PR.

omarNaifer12:
Thanks, @Philipinho ! I’ve made the PR. Please review it.

X-lem:
Thanks for fixing this. Just noticed the issue and was about to create a issue.

PR #572, opened 2024-12-20
Fix: Ensure only one emoji list appears 

omarNaifer12:
Resolve: #570 

The solution I tried is to use a shared variable to save the nodeId. I used the Atom state management system so that all components of the pages can access it. If we click on an emoji page, we store the nodeId in the Atom, and this will trigger the list of emojis to open. If we click again on the same page, we store null to close it.

Please review this approach, and if you know a better solution, let me know.

General-Docmost-Personal-MicrosoftEdge2024-12-2001-47-50-ezgif com-video-to-gif-converter

### DIFF
--- a/apps/client/src/components/ui/emoji-picker.tsx
+++ b/apps/client/src/components/ui/emoji-picker.tsx
@@ -8,6 +8,7 @@ import {
 import { useClickOutside, useDisclosure, useWindowEvent } from "@mantine/hooks";
 import { Suspense } from "react";
 const Picker = React.lazy(() => import("@emoji-mart/react"));
+import { useTranslation } from "react-i18next";
 
 export interface EmojiPickerInterface {
   onEmojiSelect: (emoji: any) => void;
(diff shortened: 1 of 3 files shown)


Philipinho:
We are using Mantine for the UI components https://mantine.dev/core/popover.
We could simplify it with Mantine's built-in `closeOnClickOutside` prop, by adding it to the Popover component.

```tsx
<Popover
      opened={opened}
      onClose={handlers.close}
      width={332}
      position="bottom"
      disabled={readOnly}
      closeOnClickOutside
    >
```

omarNaifer12:
Hi @Philipinho ,
Nice! I initially thought the clickOutside issue was a separate problem, but we can handle both at the same time. However, I tried removing opened={opened}, and it worked, but it triggered a small problem. When I open the emoji list and click the remove button or select a new emoji, the list remains open. I think handlers.close() doesn’t work because it’s controlled by opened={opened}. I also tried using closeOnClickOutside, but it didn’t work as expected.

While searching for other solutions, I found the useClickOutside hook. However, it has some limitations because the second ref for <Popover.Dropdown> is null when the opened value is false. This made it challenging for me to use the hook correctly.

Because of this, I created a custom event listener with useEffect and two refs, which allowed me to add conditions more freely. This solution works for me, but I’m not sure if there’s a better approach. Any thoughts or suggestions are welcome.

Philipinho:
I updated it to be more inline with the Mantine UI library we are using.

PR #578, opened 2024-12-21
Fix image cut in half when exporting to PDF

yashdev16:
**Description:**
This pull request fixes the issue where images are cut in half when exporting the page to PDF. The `handlePrint` function was updated to inject CSS that avoids page breaks inside images.

**Changes:**
- Updated the `handlePrint` function in `page-header-menu.tsx` to include a CSS rule that ensures images are not split across pages when printing.

**Testing:**
- Tested locally by exporting pages with images to PDF and verifying that images are not cut in half.


Philipinho:
Thank you for your PR.
This has been implemented here https://github.com/docmost/docmost/pull/583 using CSS print.

Issue #328, opened 2024-09-18
[Feature Request] Fix/Add PWA mode


robvanvolt:
Just install the web app

_Originally posted by @ados8 in https://github.com/docmost/docmost/issues/105#issuecomment-2253765560_
            
I have just come across the closed issue, that suggested to install the web app. If you do, it bookmarks the docmost website, but it does not run in PWA mode (https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps). The manifst.json or so is maybe not there or malconfigured. Especially, when you open the app with the new bookmark, the address bar is still visible, which is usually not the case in PWA mode.

dacom-dark-sun:
Confirm. This important feature and pseudo-mobile app can be reached in 24 hours.

Philipinho:
Feel free to send a PR. I will review.

Shadowfita:
Perhaps this is achieved most simply by implementing https://vite-pwa-org.netlify.app/ .

TornaxO7:
Am I understanding it correctly that this issue indirectly requests a mobile app for docmost?

arnauos:
> Am I understanding it correctly that this issue indirectly requests a mobile app for docmost?

Just PWA compliance for nice "app simulation mode" on mobile devices. No need for a published dedicated app (desirable but not a requirement)

TornaxO7:
> Just PWA compliance for nice "app simulation mode" on mobile devices. No need for a published dedicated app (desirable but not a requirement)

Hm... would you say that it's worth it to create a new issue for a dedicated app?

LinkPhoenix:
> > Just PWA compliance for nice "app simulation mode" on mobile devices. No need for a published dedicated app (desirable but not a requirement)
> 
> 
> 
> Hm... would you say that it's worth it to create a new issue for a dedicated app?

It's not a dedicated app but a Progressive Web App (PWA). Users can "install" it by pinning the website to their home screen. On Android, use Chrome's three-dot menu and select "Add to Home Screen". On iOS, tap the share button in Safari and choose "Add to Home Screen". Once pinned, it behaves like a native app, offering features such as offline mode and push notifications, without the need for app store distribution.

AimPizza:
Are there still issues with #614 or could that be merged?

cwodicka-livefront:
I see that #614 is still open, but wanted to voice that having a PWA is also desirable to me as well. Thank you!

MyWay:
+1 for this

[...]

PR #614, opened 2025-01-08
feat: progressive web app

hoiekim:
Feature request: #328

This commit makes Docmost work as a [progressive web app](https://en.wikipedia.org/wiki/Progressive_web_app). Now users can add websites that host Docmost on their home screen and it provides native-application-like experience on iOS and Android.

Currently I deployed this changes to my self-hosted Docmost app. Try adding this website to home screen on your mobile devices: 

https://doc.hoie.kim


robvanvolt:
Looks good - you could still adjust the "Backgorund" color to the background of the app, so you wouldn't have a sharp edge on the edge to the status bar. :)

aayushchugh:
@Philipinho Can you please review this?

2TAP2B:
Any updates here?

DenysMb:
> Feel free to send a PR. I will review. 

 _Originally posted by @Philipinho in [#328](https://github.com/docmost/docmost/issues/328#issuecomment-2377598884)_

There is a PR now, and it looks very simple. 
Could you review it, please?

Philipinho:
My issue with the PWA PRs so far is the many splash screen files.
If we can reduce that to the barest minimum, I am happy to revenue this PR again.

DenysMb:
@hoiekim What happens if no splash screen image are provided?

I don't have iPhone, but I would imagine that they would use a generic splash screen with the app logo if none were provided, is that right?

If so, is the splash screen really necessary?

hoiekim:
@DenysMb That's right - except there's no such thing as generic splash screen with the app logo. It'll be just blank black or white screen depending on the loading stage. I use dark mode so on my phone the app starts with black screen, then as the renderer(browser) starts it turns white until HTML elements get rendered.

Splash screen is necessary only for esthetic reasons. There's no right way to go about it but it's my understanding that having a properly color-themed splash screen gives users the impression that the app is mature, complete or well-designed. I know it sounds ambiguous and arbitrary so use your own judgment.

Although in this specific PR's situation I'd rather not insist my ego. I see that some people are uncomfortable with adding so much binary files and HTML head data. I'm proactively responding to that feedback and removed all splash screen and reverted the commits with those image files.

Thoroslives:
Any reason the merge hasnt occured?

Vito0912:
@Philipinho Can you please review this PR? It would be helpful not only on mobile but also on desktop, making it easier to use as an "dedicated app" or at least can we get an update on whether this will be merged at all?

Especially since many people seem to want this and it is not a big change to the code. The splash images appear to have been removed completely, which was one of the reasons you mentioned for not merging.

---

Btw, in the meantime, it looks like hoiekim has a workflow for pushing his fork to Docker (https://github.com/hoiekim/docmost/actions/workflows/docker-image.yml), which should point to https://hub.docker.com/r/hoie/docmost. This version probably has the PWA features (looking at the workflow), but it has not been updated for about ~1.5 months.

I also made a fork that includes this PWA (though not exactly this PR), which I plan to maintain for now as I use Docmost daily, but there are just some features missing that are stuck in PRs or as Won't fix. My fork does not have any enterprise code and includes other PRs that are not in Docmost (or issues closed as "won't fix") and that I added for my own use: https://github.com/Vito0912/forkmost. So, it is not fully backward compatible when using e.g. new blocks

Philipinho:
Soon.

reitbauer:
Hi,
are there any updates or a rough timeline for the PWA feature?
I’d really love to use Docmost, but without PWA support it’s unfortunately not very practical for my use case.
Thanks a lot for your great work on this project!

joshuaheislerde:
@Philipinho please give us some feedback, this is obviously highly anticipated by multiple of your users.

The development has basically been done and consist of only some lines of code and a simple manifest file. Why won't you just merge this PR? This will - even with a review - probably only take you about half an hour...

dougmorin:
Looking for an update on this PR. This update would be greatly appreciated. Thank you, @Philipinho!

Philipinho:
Thank you for your work on this @hoiekim.
It will be available in the next version release.

PR #633, opened 2025-01-16
Added a Web App Manifest

aminnairi:
## Rationale

This pull request adds the ability to install the application locally.

It will enable the user to "Install" the Web Application.

Installing does not mean registration a Service Worker, it means creating a Desktop Shortcut that behaves like a normal system application, removing some of the browser's decoration when launch, but keeping the installed application behavior of sending an HTTP request to fetch HTML and rendering in a browser view.

Optionally, in a near future, it can be enhanced with a Service Worker with goodies such as offline caching for instance, but is totally not mandatory, and I do not recommend it for the time being until the app reach some stability point in the development.

The big difference between a "Shortcut" and the "Installation" of a Web App is that it can be listed in the system's applications when installed, whereas a "Shortcut" only appear in the home screen for mobile users.

When installed on Gnome for GNU/Linux, it can be docked to the dock bar, can be searched for, integrates nicely with the system, and has an overall better user experience when installed than when used as a shortcut.

## Changes

I have added a couple of screenshots, I recommend adding more as the application and features are added as they are listed before the app gets installed and give a Play Store feeling of reviewing the app before installation.

I've added a link to the `index.html` file to refer to the `manifest.json` file that is read by the navigator to create the installable app.

The `manifest.json` should be pretty exhaustively filled, but there are other fields that could be added (though they are not mandatory).

A `160x160px` image has been added, unfortunately it has been upscale from the `32x32px` image, so the quality is not great, it should use the original image for better quality.


Philipinho:
Closing as PWA is now supported.

PR #666, opened 2025-01-23
feat: edit mode preference

fuscodev:
Allow users with edit role to lock/unlock pages (using _is_locked_ column in _pages_ table that is still unused)

#399 

readonly


Philipinho:
Hi @fuscodev, this looks good.
I am thinking about having  a user preference for page edit state. I think that would make page lock in its current form redundant.

My Idea for it is this. If a user has their default page edit preference set to readonly, there would be a [segmented control](https://mantine.dev/core/segmented-control/) component on each page which they can toggle. By default, each page they access will be in "Reading state" unless they click on the edit option (or shortcut e key).
image

I can find another use for the `isLocked` column in the future.

Is this something you would be willing to work on? I appreciate your contributions to the editor.
Thank you.

fuscodev:
Hi @Philipinho, in the last commit I removed the use of the `isLocked` column and added the page edit state preference.

- default: edit
- the switch in the page view doesn't change the default preference (instead of `fullWidth`)

readonly

edersong:
Does this feature have a release date?
I'm asking because it seems to be complete about two months.

MilenFrom:
Hi, any news on this?

jackeyt:
I running docmost as a docker service with version v0.20.4,but I cannot find the reading/edit button that you have mentioned。Please tell me the way  to get that?

fuscodev:
> I running docmost as a docker service with version v0.20.4,but I cannot find the reading/edit button that you have mentioned。Please tell me the way  to get that?

It's a PR, only @Philipinho can merge it

Philipinho:
The current blocker for this is me trying to figure out the best UI for this. I don’t want the editor’s top bar to become too crowded. We already have share, comment, and table of contents options there.

I am thinking we could put the toggle in the editor menu item like we did with “Full width”. 

Ideas are welcome.

alldevic:
Just in my case:

1. Sharing to menu - very rarely used
2. Replace edit/reading via icons
3. In reading mode not need to off spellchecking

image
image

jackeyt:
> The current blocker for this is me trying to figure out the best UI for this. I don’t want the editor’s top bar to become too crowded. We already have share, comment, and table of contents options there.
> 
> I am thinking we could put the toggle in the editor menu item like we did with “Full width”.
> 
> Ideas are welcome.

The way conflunce did  is that add a "edit" button  at top bar. It's a good enough.Not everyone is writing,but more forcing on reading. This way can avoid somebody touch the keyboard and input some symbol when other one is writing  by accident

MilenFrom:
Guys, the UpNote app has really nice UI for this feature and it is very convenient - at least for me.
You can see the screenshot where the button is - I'll share it here.
So basically how this is made is that this button is a toggle switching the document between two modes:
- Editing mode
- Reading mode

Can be a good idea if per workspace the user to be able to se a default mode for each new document.

SCR-20250604-kqly

fuscodev:
I think it must be in top bar (user may not remember default edit mode). Simply removing "Share" label (icon and tooltip would be enough), reducing spacing and using button switch like first commit instead of segmented control.

jackeyt:
> Guys, the UpNote app has really nice UI for this feature and it is very convenient - at least for me. You can see the screenshot where the button is - I'll share it here. So basically how this is made is that this button is a toggle switching the document between two modes:
> 
> * Editing mode
> * Reading mode
> 
> Can be a good idea if per workspace the user to be able to se a default mode for each new document.
> 
> SCR-20250604-kqly

So when can I be able to get this feature?

Hellgringo83:
Having this toggle button would be a great improvement. For my use case most users only read and some edit. So normal behaviour would be the locked mode. 
I know that this was not the initial idea so it could also be an option to set the default behaviour for an instance or even only enable the "lock" feature by a server parameter.

Philipinho:
image
The UI doesn't look too bad, though.

While the top bar seems crowded, we could release it this way and optimize later in the future.

fuscodev:
It's ok for me, but the menu button and breadcumbs won't appear on small devices if we use segmented control

Screenshot 2025-06-12 002116

code-farmer-i:
Is there any progress so far?

Philipinho:
This PR is now merged and will be available in the next release.
Thank you for your work.

Issue #313, opened 2024-09-14
[Feature Request] Support for HTTPS


RedNicStone:
Currently as far as I am aware there is no way to launch the server with TLS encryption.
This would be a beneficial feature to have in order to protect login details and other data.

Since we are running an instance of docmost on the Internet, we had to set up HTTPS encryption using an external nginx proxy which was rather complex to set up.

Adding the feature would be rather simple, see [this code section for reference](https://github.com/docmost/docmost/blob/dea9f4c063614cd884e37a3775b7d9efbc62bb36/apps/server/src/main.ts#L17). In the construction of the FastifyAdapter an optional `httpsOptions` parameter can be passed with file locations for certificates.

The location to the private and public TLS certificates could either be hard-coded or passed using an environment variable.
If no certificate is present the service could either shut down, continue running in HTTP mode, or generate a self-authored certificate (which I would prefer).

RiekertKBW:
I'm using a Caddy Reverse Proxy for HTTPS.
For now thats working great. Even if we use it just on LAN.

sabirovrinat85:
As mentioned above, just use reverse proxy, I also prefer Caddy, it's simple for most use cases and customizable when it's needed to be granularly  configured. And consider adding a geoip plugin to Caddy in order to harden security a little bit further (caddy-ipinfo-free)

RedNicStone:
Like I mentioned in the issue were currently also using a reverse proxy via nginx.

However a reverse proxy server is in no way an adequate a replacement for HTTPS support. Using a reverse proxy opens up servers to several vulnerabilities both due to user error of faulty software. A misconfiguration could lead to the vulnerable server to exposed to other system users or even the public network. Considering that docmost dosent even hash passwords client side that would be devastating.
It also increases the setup complexity and load on the server.

In an optimal configuration, if one would be to use a reverse proxy (for rate limiting and filtering for example) then the connection to the actual server would be established via HTTPS with a self-authored certificate, which is the reason I suggested this in the issue.

Considering that adding support for HTTPS is quite literally five lines of code and would massively improve security I see no reason not to add this.

lachlan-centrifuge:
> However a reverse proxy server is in no way an adequate a replacement for HTTPS support. Using a reverse proxy opens up servers to several vulnerabilities both due to user error and faulty software. A misconfiguration could lead to the vulnerable server beeing exposed to other system users or even the public network. Considering that docmost dosent even hash passwords on the client side that would be devastating.

[...]

PR #134, opened 2024-07-29
Example working helm chart

idjohnson:
I included a basic helm values as well as my example helm values (values.yaml.example)

This assumes creating a namespace and a working Redis/Valkey and PostgreSQL instance

I used
```
kubectl create ns docmost
helm install redis bitnami/redis --namespace docmost --set auth.enabled=false
helm install postgresql bitnami/postgresql --namespace docmost
```

Note, i did attempt to create the database and set the user/password (but did not work)

```
$ helm install postgresql bitnami/postgresql --namespace docmost \
  --set postgresqlUsername=docmost \
  --set postgresqlPassword=somecomplicatedpassword\
  --set postgresqlDatabase=docmost
```

I just wanted to give back if you found it at all useful

PR #678, opened 2025-01-27
[#313] feat: Add TLS support (with Caddy)

UnoYakshi:
- Add Caddy service in `docker-compose.yml` as a reverse-proxy.
- Update `.env.example` with `DOMAIN` (also rework `APP_URL`) as Caddyfile now uses `DOMAIN` and `PORT` envars.
- Remove ports for `docman` service as we only need access through 443.

Refs: #313

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -1,0 +1,4 @@
+{$DOMAIN}
+
+tls internal
+reverse_proxy docmost:{$PORT}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,8 +11,6 @@ services:
       APP_SECRET: 'REPLACE_WITH_LONG_SECRET'
       DATABASE_URL: 'postgresql://docmost:STRONG_DB_PASSWORD@db:5432/docmost?schema=public'
       REDIS_URL: 'redis://redis:6379'
-    ports:
-      - "3000:3000"
     restart: unless-stopped
     volumes:
       - docmost:/app/data/storage
@@ -33,7 +31,24 @@ services:
     volumes:
       - redis_data:/data
 
+  reverse_proxy:
+    image: caddy:latest
+    restart: unless-stopped
+    container_name: caddy_proxy
+    environment:
+      DOMAIN: ${DOMAIN}
+      PORT: ${PORT}
+    ports:
+      - 80:80
+      - 443:443
+    volumes:
+      - ./Caddyfile:/etc/caddy/Caddyfile
+      - caddy_data:/data
+      - caddy_config:/config
+
 volumes:
   docmost:
   db_data:
   redis_data:
+  caddy_data:
+  caddy_config:
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
-# your domain, e.g https://example.com
-APP_URL=http://localhost:3000
+# your domain, e.g., "example.com" or "docmost.localhost" or even "localhost"
+DOMAIN=localhost
 PORT=3000
+APP_URL=https://${DOMAIN}:${PORT}
 
 # make sure to replace this.
 APP_SECRET=REPLACE_WITH_LONG_SECRET


UnoYakshi:
I'd also like to rework PostgreSQL connection URL a little in `docker-compose.yml` and `.env.var`:

```
POSTGRES_DB=docmost
POSTGRES_PORT=5432
POSTGRES_USER=docmost
POSTGRES_PASSWORD=password
POSTGRES_HOST=localhost
DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:${POSTGRES_PORT}/${POSTGRES_DB}
```

Caddyfile
```diff
@@ -0,0 +1,4 @@
+{$DOMAIN}
+
+tls internal
```
UnoYakshi, line 3: This should most likely be deleted in production environment.

docker-compose.yml
```diff
@@ -11,8 +11,6 @@ services:
       APP_SECRET: 'REPLACE_WITH_LONG_SECRET'
       DATABASE_URL: 'postgresql://docmost:STRONG_DB_PASSWORD@db:5432/docmost?schema=public'
       REDIS_URL: 'redis://redis:6379'
-    ports:
-      - "3000:3000"
```
UnoYakshi, line 15: With a reverse-proxy we won't 3000 port need to open the host machine.

UnoYakshi:
I'd also prefer to separate `.env.example` to:
- `.env/app.env`
- `.env/db.env`
- `.env/email.env`
- `.env/aws.env`

and utilize those in `docker-compose.yml`:
```yml
version: '3'

services:

  docmost:
    image: docmost/docmost:latest
    container_name: docmost
    depends_on:
      - db
      - redis    
    env_file:
      - path: .env/app.env
        required: true
      - path: .env/db.env
        required: true
      - path: .env/email.env
        required: false
      - path: .env/aws.env
        required: false
    restart: unless-stopped
    volumes:
      - docmost:/app/data/storage

  db:
    image: postgres:16-alpine
    env_file:
      - path: .env/db.env
        required: true
    restart: unless-stopped
    volumes:
      - db_data:/var/lib/postgresql/data

  redis:
    image: redis:7.2-alpine
    restart: unless-stopped
    volumes:
      - redis_data:/data

  reverse_proxy:
    image: caddy:latest
    restart: unless-stopped
    container_name: caddy_proxy
    env_file:
      - path: .env/app.env
        required: true
    ports:
      - 80:80
      - 443:443
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile
      - caddy_data:/data
      - caddy_config:/config

volumes:
  docmost:
  db_data:
  redis_data:
  caddy_data:
  caddy_config:
```

UnoYakshi:
@Philipinho, any update?

ATofighi:
I don't like this PR :D. I believe adding a reverse proxy (like Caddy) and enforcing HTTPS to the quick-start Docker Compose introduces unnecessary complexity for initial testing. The default docker-compose.yml should stay minimal and simple. Production-grade concerns, such as HTTPS enforcement and reverse proxies, are better handled with separate documentation or optional configurations.

UnoYakshi:
> I believe adding a reverse proxy (like Caddy) and enforcing HTTPS to the quick-start Docker Compose introduces unnecessary complexity for initial testing.

I don't really see a problem with HTTPS being turned on by default. Can you elaborate?

Also, no one forces you to `up` all the services at once. Just run `docker compose up docmost` if you don't want the proxy. But I don't really see the benefit of running docmost in Docker-environment for development ― from what I recall, setting up debugging within IDE+Docker is (or was) not straightforward, to say the least.

ATofighi:
> > I believe adding a reverse proxy (like Caddy) and enforcing HTTPS to the quick-start Docker Compose introduces unnecessary complexity for initial testing.
> 
> I don't really see a problem with HTTPS being turned on by default. Can you elaborate?
> 
> Also, no one forces you to `up` all the services at once. Just run `docker compose up docmost` if you don't want the proxy. But I don't really see the benefit of running docmost in Docker-environment for development ― from what I recall, setting up debugging within IDE+Docker is (or was) not straightforward, to say the least.

I’m not referring specifically to local development; I’m talking about the initial onboarding experience—when someone just wants to quickly spin up Docmost to see if it meets their needs. A minimal `docker-compose.yml` lowers the barrier to entry for these first-time tests.

In my experience, most open-source projects don’t include a reverse proxy in their default Docker Compose file. Instead, they typically ship only the required dependencies and place proxy or SSL configurations in a separate file or documentation. I double check two other projects again:
- WordPress ( https://github.com/docker-library/docs/tree/master/wordpress#-via-docker-compose-or-docker-stack-deploy ):  
  The default `docker-compose.yml` only exposes port 80. For SSL, an additional configuration is required.  
- Sentry ( https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml#L454C11-L454C22 ):  
  Even though Sentry is quite complex, the default setup only exposes port 80, leaving SSL to be handled by the user via a reverse proxy.

I think that adding Caddy as a default dependency in `docker-compose.yml` is unnecessary and creates additional complexity for first-time users.

UnoYakshi:
> > > I believe adding a reverse proxy (like Caddy) and enforcing HTTPS to the quick-start Docker Compose introduces unnecessary complexity for initial testing.
> > 
> > 
> > I don't really see a problem with HTTPS being turned on by default. Can you elaborate?
> > Also, no one forces you to `up` all the services at once. Just run `docker compose up docmost` if you don't want the proxy. But I don't really see the benefit of running docmost in Docker-environment for development ― from what I recall, setting up debugging within IDE+Docker is (or was) not straightforward, to say the least.
> 
> I’m not referring specifically to local development; I’m talking about the initial onboarding experience—when someone just wants to quickly spin up Docmost to see if it meets their needs. A minimal `docker-compose.yml` lowers the barrier to entry for these first-time tests.
> 
> In my experience, most open-source projects don’t include a reverse proxy with SSL in their default Docker Compose file. Instead, they typically ship only the required dependencies and place proxy or SSL configurations in a separate file or documentation. I double check two other projects again:
> 
>     * WordPress ( https://github.com/docker-library/docs/tree/master/wordpress#-via-docker-compose-or-docker-stack-deploy ):
>       The default `docker-compose.yml` only exposes port 80. For SSL, an additional configuration is required.
> 
>     * Sentry ( https://github.com/getsentry/self-hosted/blob/master/docker-compose.yml#L454C11-L454C22 ):
>       Even though Sentry is quite complex, the default setup only exposes port 80, leaving SSL to be handled by the user via a reverse proxy. (they have reverse proxy in their default docker-compose, but not because of ssl, they use it as a gateway of other services)
> 
> 
> I think that adding Caddy as a default dependency in `docker-compose.yml` is unnecessary and creates additional complexity for first-time users.

Why does running `docker compose up` with a Caddy-service add more complexity?

You don't need to set it up anyhow [until you need to tweak it]. If you have another reverse-proxy (most likely, it's Nginx and is compatible with Caddy, or K8S/Helm/Terraform/etc. which is another beast), you just exclude Caddy as I mentioned. if you have your infrastructure setup already, you can definitely solve this "problem".


> In my experience, most open-source projects don’t include a reverse proxy with SSL in their default Docker Compose file.

1. First of all, what other projects do en masse is not a valid argument by itself. It should have reasons. You didn't really provide those, if you ask me.
Not to mention I'd love more projects to have SSL configured automatically for the very reason of testing them in staging or home environment in the first place.

2. I presume not bringing a set-up reverse-proxy is, at most, a deliberate solution to avoid a potential scope of development (which I can agree on sometimes; `it_depends.png`). By that, they achieve _infrastructural agnosticism_.

3. Before Caddy, most (if not all) "SSL pipelines" required extra setting up: certificates, secure settings for the server, etc. Caddy helps mitigating this extra work, bringing easy-to-SSL (automatic!) feature most projects lack. Like Sentry and Wordpress, that you mentioned.
3.1. https://developer.wordpress.org/advanced-administration/security/https/#using-a-reverse-proxy ― BTW, WP explicitly state that people should run it behind HTTPS. I presume, the issue is with SSL certificate setup.
> WordPress is fully compatible with HTTPS when **an TLS / SSL certificate is installed and available** for the web server to use. Support for HTTPS is strongly recommended to help maintain the security of both WordPress logins and site visitors.
3.2. Considering how widespread WP is, it'd be risky to add Caddy's automatic certificate issuing, etc. It simply could break way too many services.
Docmost, however, can afford a move like that.

4. Sentry and WordPress aren't the best examples to compare to.
4.1. Sentry, by design, is a software complementary to some other software. I can't really imagine a startup company's first deployment being Sentry. Hence, what is the reason to have SSL deployment configurations for such software? But I'd still include it, though, exactly for the newcomers.
4.2. WordPress is so integrated into our society, it doesn't really need an automatic SSL in Docker Compose YAML. Plenty of services (VPS, website hosting) provide HTTPS support already, alongside of serving your WP with 1 click.
4.2.1. I can clearly imagine WordPress Foundation being paid to NOT include automatic SSL/TSL. I remind you, some hosting services still ask for money for HTTPS setup.

5. As I mentioned, Docmost is different from the mentioned Sentry and WP. It's not well-known (yet) and is stand-alone software. So I (and other people) don't see a problem of including HTTPS turned on by default.
The whole point of [the 313 issue](https://github.com/docmost/docmost/issues/313) and this very PR is to bring **HTTPS by default** with **no extra configurations** needed. Because some members of Docmost's **community already want and do** that.
We don't look at WordPress and say, "Hm-m, they don't have SSL/TSL in their Docker orchestration setup ― we shouldn't have it either!"

6. BTW, have you pulled this branch/PR and tried it out? Because I don't see any "onboarding" issues here as it is the same `docker compose up`. It won't eat 2G of RAM and won't ask for extra input. Quite the opposite, it helps with a more proper-ish deployment.

UnoYakshi:
Also, @ATofighi, all this discussion should be in the issue, not in the PR.

UnoYakshi:
@Philipinho, any chance to review the PR?

Philipinho:
Hi @UnoYakshi, thank you for your contribution.

While it seems valuable, as previously stated here https://github.com/docmost/docmost/pull/134#issuecomment-2405334908, I prefer to not put infrastructural code in the main repo.
In the future, I will consider creating a separate repo for setups like this.

The purpose of the base docker compose file is to give users a starting point. They are then free to explore with whatever reverse proxy setup or docker enhancements which works for them.
I hope you understand.

UnoYakshi:
@Philipinho, thanks for the reply.

# Include `*caddy.yml`
By any chance, can we, perhaps, move the Caddy setup in a separate Docker Compose configuration (e.g., `docker-compose.caddy.yml`) and make it optional via [include](https://docs.docker.com/reference/compose-file/include/)?
As I mentioned before, I do think more projects should have «one-click production setup». It saves time and [mental] energy; helps with early adopters. For some, it may even be the turning point.

If your decision is final, should I close the PR or remain it open? There is no option to «freeze it until X» or something.

# Separate Repository?
I also wonder how do you see a separate repository? Is approach going to be:
- an **opinionated** ― a specific set of tools and configurations ― what I've seen in most commercial projects so far,
- or **agnostic** ― only provide minimal setup with "swapable architecture" ― not sure if it's possible/viable/worth the time,
- or **checklist** ― choose this or that for whatever case you might need ― tons of work, but ?

UnoYakshi:
@Philipinho, moved my reply back here.


# > I prefer to not put infrastructural code in the main repo.

I don't think we understand it the same way. Isn't `docker-compose.yml` already a part of the infrastructure? And if we have one infrastructural (for now, minimal) thingy, why not include the other?

The latter "why not include the other" can lead the project into a mess, however, if the borderline is not clearly defined. And you somewhat define this line…

> The purpose of the base docker compose file is to give users a starting point. They are then free to explore with whatever reverse proxy setup or docker enhancements which works for them.

Let's sit on it. I'd like to focus on _free to explore_ for a moment as I see only 3 main cases here:
- A[n] person/team/organization already know what they are doing and they already have their infrastructure set-up, with SSO/LDAP/etc. But they won't explore, they will want to properly setup ASAP. In that case, the minimal Docker Compose YAML is a good thing.
- A newcomer who's probably not that good with Docker Compose and the whole infrastructural shebang (we all have been there). So the reverse-proxy being setup is a good example for him or her and will actually help to explore, unlike not having this option, IMO.
- The last case is the intermediate between the first two. And it is the case for some of the community members, including myself. We want a simple setup with essentials (and I consider SSL/TSL an essential in modern Web) working out-of-box. We do know how to make it work via Caddy/Nginx, Traefik, clouding stuff, but that's a boring routine. And if we can evade it w/o [negative] repercussions, I bet we'll try to.

So we have 3 groups. And only a single case doesn't want any extra setup. Others 2 either don't mind it or will greatly benefit from it. How about a compromise then (next section)?


# Solution: `include *caddy.yml`
If we want to keep it minimal (your point) and to include a bit more infrastructure (my point). can we, perhaps, move the Caddy setup in a separate Docker Compose configuration (e.g., `infrastructure/docker-compose.caddy.yml`) and make it optional via [include](https://docs.docker.com/reference/compose-file/include/)? I'd even see commented-out included service be a viable solution here.
As I mentioned before, I do think more projects should have «one-click production setup». It saves time and [mental] energy; helps with early adopters. For some, it may even be the turning point.

If your decision is final, should I close the PR or remain it open? There is no option to «freeze it until X» or something.

# Separate Repository?
I also wonder how do you see a separate repository? Is approach going to be:
- an **opinionated** ― a specific set of tools and configurations ― what I've seen in most commercial projects so far,
- or **agnostic** ― only provide minimal setup with "swapable architecture" ― not sure if it's possible/viable/worth the time,
- or **checklist** ― choose this or that for whatever case you might need ― tons of work and probably isn't worth it before the 1.0 release
?

Philipinho:
Closing as I believe this should belong to docs or a separate guides repo.
I am not against the idea of it, just don't want infrastructure code like to this to be on the main repo.

Issue #677, opened 2025-01-27
Numbered List or To-Do List Not Working Inside Tables


jay311992:
When using Docmost, I am unable to add numbered lists or to-do lists inside table cells. The functionality either does not work as expected or is unavailable.

Further testing allow shows any formatting option is not populated in the table cell including headings, etc. 

**Steps to Reproduce:**

Create a new table in Docmost.

Click inside a table cell and attempt to add a numbered list or to-do list.

Observe that the list formatting does not apply or the feature is non-functional.

**Expected Behavior:**
The numbered list or to-do list should be properly formatted and functional within table cells, similar to how it works outside of tables.

**Actual Behavior:**
The list formatting does not work, or the feature is unavailable inside table cells.

**Environment:** 

Version: [v0.7.0](https://github.com/docmost/docmost/releases)
Browser: Edge
Deployed through Docker Container

mustafa-yilmaz:
I guess this request is related to the one I created at #655 ?

Philipinho:
Supported in latest versions.

PR #679, opened 2025-01-27
fix: allow every content block in table cell

fuscodev:
#677

tablecell

### DIFF
--- a/apps/client/src/features/editor/extensions/extensions.ts
+++ b/apps/client/src/features/editor/extensions/extensions.ts
@@ -139,7 +139,9 @@ export const mainExtensions = [
     allowTableNodeSelection: true,
   }),
   TableRow,
-  TableCell,
+  TableCell.extend({
+    content: 'block+',
+  }),
   TableHeader,
   MathInline.configure({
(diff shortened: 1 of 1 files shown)


mustafa-yilmaz:
Just a humble recommendation.

For the user experience (ux) of this feature, the "**column type**" feature of "**Microsoft Loop**" could be used as a base inspiration as I mentioned at #655 .  

Thanks.

Philipinho:
@fuscodev ~The table is limited to better support Markdown exports. I don't think we will be extending it at this time.~
Edit: supported in latest versions.

@mustafa-yilmaz, for now, we don't have plans to work on such. If we eventually do, It would be a different table node.

Vito0912:
I understand that there should be good Markdown support. 

However, why is it allowed in the header but not in the rows of the table? This seems like a significant missing feature. Even if this is missing because it's supposed to be "better Markdown," support, why is it allowed in the header? If people want to use it, why is it not permitted?

Especially since there are ways to get it to work. Just more inconvinient. Therefore even less understand these set limitations

PR #691, opened 2025-01-30
fix: default language on invitation signup

Philipinho:
* Set default language on invitation signup
* Default language selection to en-US if locale is undefined (Client)

### DIFF
--- a/apps/server/src/database/repos/user/user.repo.ts
+++ b/apps/server/src/database/repos/user/user.repo.ts
@@ -99,7 +99,8 @@ export class UserRepo {
     trx?: KyselyTransaction,
   ): Promise<User> {
     const user: InsertableUser = {
-      name: insertableUser.name || insertableUser.email.toLowerCase(),
+      name:
+        insertableUser.name || insertableUser.email.split('@')[1].toLowerCase(),
       email: insertableUser.email.toLowerCase(),
       password: await hashPassword(insertableUser.password),
       locale: 'en-US',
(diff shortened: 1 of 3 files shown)


apps/server/src/database/repos/user/user.repo.ts
```diff
@@ -99,7 +99,8 @@ export class UserRepo {
     trx?: KyselyTransaction,
   ): Promise<User> {
     const user: InsertableUser = {
-      name: insertableUser.name || insertableUser.email.toLowerCase(),
+      name:
+        insertableUser.name || insertableUser.email.split('@')[1].toLowerCase(),
```
ATofighi, line 103: is this line ok?
for exmpale if user email was name@gmail.com and there is no name, then name will set to gmail.com

Philipinho: Fixed https://github.com/docmost/docmost/commit/0f74f032641a4dd07687ee60facfc850e9da62cc. Thanks for catching.

apps/server/src/database/repos/user/user.repo.ts
```diff
@@ -110,7 +111,7 @@ export class UserRepo {
     const db = dbOrTx(this.db, trx);
     return db
       .insertInto('users')
-      .values(user)
+      .values({ ...insertableUser, ...user })
```
alldevic, line 114: ERROR [ExceptionsHandler] column "workspace_name" of relation "users" does not exist on first start creating workspace

Philipinho: Fixed.

Issue #662, opened 2025-01-23
[Bug] Sidebar Spaces change causes tab to close once switching to another Space.


abramter:
Hi,

When I'm in any of the Spaces I have made and try and switch to another space from the left sidebar pulldown menu, when I click to another Space, the page closes (chrome tab closes).

ijustyce:
Me too, but this issue will be resolved after create some documents under this space.

hermda02:
This happens for us too, even if there are some documents under each space.

fuscodev:
I think I fix it in PR #692
There was a `close()` method without `useDisclosure()` hook in `switch-space.tsx`

Philipinho:
The fix has been merged. Will be available in the next release.

PR #692, opened 2025-01-30
fix: switch space bug

fuscodev:
#662 

| Before fix| After fix|
|--------|--------|
| change-space-old| change-space-new|


Philipinho:
Good catch. The undefined `close()` fell back to `window.close()`, which closes the tab.

PR #702, opened 2025-02-06
fix: enforce 32-character minimum length for APP_SECRET

Philipinho:
For better security, we are enforcing a minimum of 32-character length for the `APP_SECRET` env variable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@
 APP_URL=http://localhost:3000
 PORT=3000
 
-# make sure to replace this.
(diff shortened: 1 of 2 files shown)


kwiniarski97:
Is there some migration guideline for people who have set up docmost previously with shorter secret?

Philipinho:
Hi @kwiniarski97 just generate a new secret and everything will work fine.
You can use this command: `openssl rand -hex 32`.

PR #720, opened 2025-02-11
Fix: Close Popover When Clicking Outside in SwitchSpace Component

auxa-m45:
This PR fixes an issue where the popover in the SwitchSpace component does not close when clicking outside.

Since SwitchSpace is a controlled component, the state needs to be toggled via onChange. This change ensures that the popover properly closes when expected.

before
after

Reference: [mantinedev/mantine#7019](https://github.com/mantinedev/mantine/issues/7019)


Philipinho:

PR #733, opened 2025-02-16
fixes infinite rendering when selecting member to add to space

hoiekim:
Just noticed that `MultiMemberSelect` components renders infinitely when you type in a name to select a member to add to a space. The root cause is because `useEffect` in this component is depend on a state variable `data` which is updated every time running the callback of `useEffect`. It can be fixed by removing the use of `data` variable by replacing it with the argument passed in `setData`. See the code diff for details.

### DIFF
--- a/apps/client/src/features/space/components/add-space-members-modal.tsx
+++ b/apps/client/src/features/space/components/add-space-members-modal.tsx
@@ -19,14 +19,6 @@ export default function AddSpaceMembersModal({
   const [role, setRole] = useState<string>(SpaceRole.WRITER);
   const addSpaceMemberMutation = useAddSpaceMemberMutation();
 
-  const handleMultiSelectChange = (value: string[]) => {
-    setMemberIds(value);
-  };
-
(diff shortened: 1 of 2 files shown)


apps/client/src/features/space/components/add-space-members-modal.tsx
```diff
@@ -19,14 +19,6 @@ export default function AddSpaceMembersModal({
   const [role, setRole] = useState<string>(SpaceRole.WRITER);
   const addSpaceMemberMutation = useAddSpaceMemberMutation();
 
-  const handleMultiSelectChange = (value: string[]) => {
-    setMemberIds(value);
-  };
```
hoiekim, line 24: Minor performance improvement: 

Removed this because `handleMultiSelectChange` is basically an alias of `setMemberIds` but doesn't do anything else. Passing `setMemberIds` directly to child components doesn't make them re-rendered when the parent is rendered. Same applies to `handleRoleSelection`

Issue #203, opened 2024-08-24
[Feature request] Add publish page


kbkozlev:
It would be nice to be able to "publsih" pages so that they are publicly available through a link that does not require you to login, or atleast have a toggle between view and edit mode to make sure that you don't accidentally edit the page when reading the documentation.

stnby:
At least view mode would be nice first. I keep accidentally moving stuff around with touchscreen.

johhnry:
This would be so cool! I have a use case where I need to share a single public link to new artists coming into my studio. I would like to share a documentation page without having them to connect.

dankd123:
A view/edit toggle would be really handy. Public pages are covered under issue #23

dalekirkwood:
public pages is one of the most in-demand features I see requested on Reddit

lodgmar:
> It would be nice to be able to "publsih" pages so that they are publicly available through a link that does not require you to login, or atleast have a toggle between view and edit mode to make sure that you don't accidentally edit the page when reading the documentation.

I think the view/edit toggle is the simplest solution in order to avoid accidental changes.

roablep:
+1 to this. 
I'd LOVE to have a public view

Raytle:
Bump!

mustafa-yilmaz:
I guess this request is related to the one at #608 ?

Shousokon:
+1

Reetryyy:
I also would like to see this feature in Docmos!

gitlipf:
+1

Oswald522:
+1

Philipinho:
Starting from [v0.20.0](https://github.com/docmost/docmost/releases/tag/v0.20.0), you can now publicly share pages.

Issue #23, opened 2024-06-30
Public Wiki


ItsNoted:
Will there be options to make the wiki public so login is not required? Or certain wiki pages? Love the project thank you!

Also, why is your documentation using Docusaurus? It's a great opportunity to showcase your own app. Seems a bit odd, I assume because of the very issue I am asking about? Because it's a closed, private wiki?

Philipinho:
I would say for now, Docmost is more useful for internal and private wiki/docs.

Ideally, I would have started with a public wiki. It was in my initial development goals. I realized it was better to wait, understand the usage and figure out how best to expose the wiki, permissions, URL structure and some other factors to put into consideration.
This feature will come but just not yet.
Sharing individual pages will come first.

Unrelated: I like your site [noted.lol](https://noted.lol), I have come across it a lot.

sepbin:
I'm looking forward to this feature, if released it will be my main productivity tool

GordonHuangYong:
public work space. not single page

dankd123:
This is the top item on my wishlist. Better yet, I would love to be able to restrict even specific sections of a page as password-restricted while leaving the rest of the page publicly-accessible. This way we can write one page for a support issue, and simply redact the internal info from the public-facing version of the page.

KPGTB:
Any updates?

HelloGithubUniverse:
Any updates on this? Would be a great feature!

nikkoxgonzales:
Bump on this please! :)

HelloGithubUniverse:
Just to add: bookstack solved this with the role „guest“. Every not registered user is a guest.

tylerdurden4285:
+1 please.

llmoe:
+1

suyang1023:
+1

fadlee:
+1

Joonas12334:
+1

simper:
+1

HelixGem:
+1

SirLouen:
+1

kovai:
+1

kovai:
> I've been using this patch that adds this functionality. [#764](https://github.com/docmost/docmost/pull/764) So far, so good.

Thanks a lot!

Oswald522:
I'm looking forward to this feature, if released it will be my main productivity tool

Joonas12334:
+1 so far i love the writing and organizational tools. Just add public links and that's it!

colletda:
+1

[...]

PR #764, opened 2025-02-20
feat: publish spaces and let anonymous users access to pages in published spaces

hoiekim:
See demo: https://doc.hoie.kim/share/s/public/p/product-specification-T76LEgs2eq

This might resolve feature request #203 and #23

Note that there's a subtle difference: some requested a feature to publish pages while others wanted a feature to publish the whole spaces. This PR adds feature to publish spaces and all pages in published spaces will be accessible by anonymous users.

~~User interface changes:~~
* ~~`Publish space` button will be added next to `Add space members` button in `Space settings` -> `Members` modal.~~
* ~~Anonymous users won't have user icon and dropdown menu on the upper right hand corner~~
* ~~Anonymous users won't have space list dropdown or space setting menu in the side bar~~
* ~~Anonymous users won't have user preference update buttons in the page kebab menu~~

~~Authentication mechanism changes:~~
* ~~`JwtAuthGuard` won't throw exception if the "referer" header in the request implies that the request is from a published space. Due to this change I added anonymous user handling mechanisms for a few controllers~~
*  ~~When the request is from a published space, `AuthUser` decorator will return anonymous user object. This object represents a hypothetical user data that is actually not stored in DB. This user object is useful when resolving `workspaceAbility` and `spaceAbility`.~~

## Updates on March 17, 2025

User interface changes:
* `Publish space` button will be added next to `Add space members` button in `Space settings` -> `Members` modal.
* Published spaces have "Copy shared link" button in page header menu and page tree menu.
* Shared pages and spaces are accessible via URL starts with `/share`, such as `/share/s/myspace/p/mypage`, etc.

space-setting-modal-screen
not-shared-page-screen
shared-page-screen

Both backend and frontend endpoints for shared space are separated from the existing ones and have separate authorization process. To do this I created `ShareController` that has similar methods with `SpaceController` and `PageController` but does not have `JwtAuthGuard`. Instead it authorizes requests from unidentified users if the space in question is a "published" space.

### DIFF
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -7,7 +7,7 @@ import AccountSettings from "@/pages/settings/account/account-settings";
 import WorkspaceMembers from "@/pages/settings/workspace/workspace-members";
 import WorkspaceSettings from "@/pages/settings/workspace/workspace-settings";
 import Groups from "@/pages/settings/group/groups";
-import GroupInfo from "./pages/settings/group/group-info";
+import GroupInfo from "@/pages/settings/group/group-info";
 import Spaces from "@/pages/settings/space/spaces.tsx";
 import { Error404 } from "@/components/ui/error-404.tsx";
 import AccountPreferences from "@/pages/settings/account/account-preferences.tsx";
@@ -17,7 +17,7 @@ import Layout from "@/components/layouts/global/layout.tsx";
 import { ErrorBoundary } from "react-error-boundary";
 import InviteSignup from "@/pages/auth/invite-signup.tsx";
 import ForgotPassword from "@/pages/auth/forgot-password.tsx";
-import PasswordReset from "./pages/auth/password-reset";
+import PasswordReset from "@/pages/auth/password-reset";
 import Billing from "@/ee/billing/pages/billing.tsx";
 import CloudLogin from "@/ee/pages/cloud-login.tsx";
 import CreateWorkspace from "@/ee/pages/create-workspace.tsx";
@@ -26,6 +26,9 @@ import { useTranslation } from "react-i18next";
 import Security from "@/ee/security/pages/security.tsx";
 import License from "@/ee/licence/pages/license.tsx";
 import { useRedirectToCloudSelect } from "@/ee/hooks/use-redirect-to-cloud-select.tsx";
+import UserAgnosticLayout from "@/components/layouts/global/user-agnostic-layout";
+import SharedPage from "@/pages/page/shared-page";
+import SharedSpaceHome from "@/pages/space/shared-space-home";
 
 export default function App() {
   const { t } = useTranslation();
@@ -52,6 +55,21 @@ export default function App() {
         )}
 
         <Route path={"/p/:pageSlug"} element={<PageRedirect />} />
+
+        <Route path={"/share"} element={<UserAgnosticLayout />}>
+          <Route path={"s/:spaceSlug"} element={<SharedSpaceHome />} />
+          <Route
+            path={"s/:spaceSlug/p/:pageSlug"}
+            element={
+              <ErrorBoundary
+                fallback={<>{t("Failed to load page. An error occurred.")}</>}
+              >
+                <SharedPage />
+              </ErrorBoundary>
+            }
+          />
+          <Route path={"p/:pageSlug"} element={<PageRedirect />} />
+        </Route>
 
         <Route element={<Layout />}>
           <Route path={"/home"} element={<Home />} />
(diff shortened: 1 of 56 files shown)


hoiekim:
@Philipinho I'll appreciate your feedback on this

Philipinho:
Hi @hoiekim, I have had a look at your work.
I would say, it looks mostly good, but I do not wish to open the core dashboard in this manner, as it would open us to potential security issues. It will be very difficult to manage permissions consistently.

I am considering the possibilities of public sharing, and I lean towards it having it's own namespace. 
We can expose sharing through a set of routes, e.g `mywiki.com/share/xzy`. This way, we don't have to worry about trying to manage who has what access to what in both the UI and the Server.

hoiekim:
Makes sense to me. I'll update as I get some time to work on it

dcgsteve:
@hoiekim @Philipinho - from a pure user perspective, I love the approach you've taken here - allocating a space to be published is definitely a good move in my opinion.  Publishing at page level is very awkward to manage and leaves the potential for pages to still be public that you've forgotten about. Thanks for your hard work both!

SirLouen:
Despite its not even merged into master, I'm testing this in a production environment and I'm amazed with the feature so far. ✅ So far my tests are passing

hoiekim:
Noticed that the compile fails after merging with the latest changes so rebased on main branch and resolved the error. All looks good now.

SirLouen:
It would be great if single pages could be shared (not the whole workspace). I thought that the last patch added this.

SirLouen:
I've found a weird bug:
Sometimes, not sure why, it shows the info for a second and then redirects automatically 
For example with this first link it redirects:
https://test.wp.voz.cat/share/s/ideas/p/testing-workflow-ideas-U9UoGrVDLP
But with these two others, it doesnt:
https://test.wp.voz.cat/share/s/ideas/p/unofficial-mentorship-program-ePhW1o3sum
https://test.wp.voz.cat/share/s/ideas/p/about-meetings-protocol-Xl7h0JnhDf
Both for the same Workspace and I can't see a difference? (maybe it only affects the first doc in the list?)

SirLouen:
@hoiekim 

I've found a weird bug:
Sometimes, not sure why, it shows the info for a second and then redirects automatically
For example with this first link it redirects:
https://test.wp.voz.cat/share/s/ideas/p/testing-workflow-ideas-U9UoGrVDLP
But with these two others, it doesnt:
https://test.wp.voz.cat/share/s/ideas/p/unofficial-mentorship-program-ePhW1o3sum
https://test.wp.voz.cat/share/s/ideas/p/about-meetings-protocol-Xl7h0JnhDf
Both for the same Workspace and I can't see a difference? (maybe it only affects the first doc in the list?)

hoiekim:
@SirLouen Thanks for your efforts with testing. I experienced the same when I clicked the links you provided.

I think it's related with the linked pages inside the document. I suspect that the auth-required API endpoint is called while resolving the link. I'll fix this some time soon.

Regarding the feature to publish single page, I think we can leave that as a separate feature and can be added after this PR is merged.

hoiekim:
Fixed the bug that certain extension elements in the editor caused re-direction to login page. I created another extension set called `readonlyExtensions`, which has shared version of embedded elements like linked page, image, draw.io, excalidraw and etc.

Philipinho:
@hoiekim I really appreciate the time and effort you’ve put into this. 
While the outcome looks desirable, there are too many code changes I do not agree with, which would make maintenance a burden. For example, the duplication of code in recent changes and the number of files impacted.
Tbh, I don't think It would make it to the codebase in it's current form.

I am sorry that I am not able to give much feedback on this. I also want to be transparent that I am not yet ready to commit to the implementation of this feature. When I decide to, I am sure there would be much to learn from your implementation.

I hope this does not come off as rude or discouraging in contributing in other ways.

hoiekim:
Honestly I'm disappointed but I'll take your words. Hope you'll find a better way around on this

SirLouen:
@hoiekim this is not an optional feature for a system of this nature. I will be probably be forking for now your advances in this regard. Prob a ton of people will be appreciating all the polishes around this feature that you have been doing.

If I were @Philipinho I would have added this permanently into the trunk version until I had the time to code review it and back port it into an stable release. Probably he could take big parts, maybe not all, of the code for this purpose.

Also @Philipinho if you have a guy that is PRing you, why you don't give some tips on which direction from your point of view would be preferred to facilitate the integration this? Also if you check in issues this is not any close the number one requested feature by everyone. You have the helpers, the demand, you have everything in your hands, it's strange you are not taking the opportunity to make a massive step in the software by focusing in such minor feature.

Philipinho:
Public page sharing is being worked ob. PR: https://github.com/docmost/docmost/pull/1012.
Closing.

PS: @hoiekim, I truly appreciate the time and effort you put in this PR. Thank you.

PR #769, opened 2025-02-20
fix: table header in markdown export

fuscodev:
Removing colgroup tags in html string before exporting in markdown

```
From isHeadingRow() function in @joplin/turndown-plugin-gfm :
// A tr is a heading row if:
// - the parent is a THEAD
// - or if its the first child of the TABLE or the first TBODY (possibly
//   following a blank THEAD)
// - and every cell is a TH
````
before -> after
Screenshot 2025-02-20 201533


Philipinho:
Thanks for your contribution.

PR #773, opened 2025-02-21
fix: accept invitation password hashing

h4rb4rd:
The password is hashed twice in acceptInvitation func and in insertUser func so we get an error when we compare the password hash when we login after invitation.


Philipinho:
Thanks for catching.  Just released [v0.8.2](https://github.com/docmost/docmost/releases/tag/v0.8.2).

Issue #534, opened 2024-12-05
Feature Request: Make Redis family Configurable via Environment Variable


iagocavalcante:
#### Context

Currently, the Redis configuration in the application dont able to pass family: 6 to redis

```typescript
new Redis({
  host: this.redisConfig.host,
  port: this.redisConfig.port,
  options: {
    password: this.redisConfig.password,
    db: this.redisConfig.db,
    retryStrategy: createRetryStrategy(),
    family: 6,
  },
});
```

#### Proposed Solution

Introduce an environment variable to configure the `family` option dynamically. For example:

- If `REDIS_FAMILY` is set, its value will be used (`REDIS_FAMILY=6` for IPv6 or `REDIS_FAMILY=4` for IPv4).
- If `REDIS_FAMILY` is unset, omit the `family` option altogether and allow Redis to determine the default based on the network.

#### Implementation Example

Update the Redis initialization to include a conditional assignment for `family`:

```typescript
new Redis({
  host: this.redisConfig.host,
  port: this.redisConfig.port,
  options: {
    password: this.redisConfig.password,
    db: this.redisConfig.db,
    retryStrategy: createRetryStrategy(),
    ...(process.env.REDIS_FAMILY ? { family: parseInt(process.env.REDIS_FAMILY, 10) } : {}),
  },
});
```

#### Acceptance Criteria

1. The application should honor the `REDIS_FAMILY` environment variable, defaulting to its absence if not specified.
2. Test cases should cover scenarios for both IPv6 (`REDIS_FAMILY=6`) and IPv4 (`REDIS_FAMILY=4`).
3. Documentation updates to describe how to configure `REDIS_FAMILY` for different network environments.

#### Benefits

This enhancement ensures better flexibility and compatibility for various deployment setups, improving the application's configurability and portability.

ju-li:
This is important for deploying on services such as Railway or Fly.io which use IPV6 for internal networking. 

If you connect to Redis using their internal networking (and IPV6), data transfers are free. 
Connecting using public networking (IPV4) is possible for now but incurs data transfer fees.

iagocavalcante:
> This is important for deploying on services such as Railway or Fly.io which use IPV6 for internal networking.
> 
> If you connect to Redis using their internal networking (and IPV6), data transfers are free. Connecting using public networking (IPV4) is possible for now but incurs data transfer fees.

Yeah, I do this changes on my forked repo, i will do a propose for this issue here

ju-li:
[...]

PR #807, opened 2025-02-26
feat: adding family 6 in uri to configure for both 4 and 6

iagocavalcante:
closes https://github.com/docmost/docmost/issues/534#issuecomment-2686172185


Philipinho:
Thanks for your contribution.

PR #829, opened 2025-03-04
Update docker-compose.yml

DioCrafts:
the Attribute 'version' is obsolete, removing it


Philipinho:
I will consider removing it in the future. For now, users can do so manually.
Thank you.

PR #835, opened 2025-03-05
improve Dockerfile security and reduce image size

DioCrafts:
# Dockerfile Optimizations

This PR improves the Dockerfile with several optimizations focused on security, image size, and build efficiency:

## Security Improvements
- Added system package updates (`apk update && apk upgrade`) to ensure the base image has the latest security patches
- Retained the existing `--no-cache` flag for `apk` which prevents storing the package cache

## Size Optimizations
- Added cache cleaning after dependency installation (`pnpm cache clean --force`)
- Removed temporary files (`rm -rf /tmp/*`) to reduce image size
- Combined RUN commands in the builder stage to reduce Docker layers

## Build Efficiency
- Kept the multi-stage build approach which was already well-implemented
- Maintained `--frozen-lockfile` to ensure dependency consistency

These changes maintain the existing functionality while producing a more secure and lightweight Docker image, which should lead to faster deployments and a smaller attack surface.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,13 +7,14 @@ WORKDIR /app
 
 COPY . .
 
-RUN npm install -g pnpm
-RUN pnpm install --frozen-lockfile
-RUN pnpm build
+RUN npm install -g pnpm && \
+    pnpm install --frozen-lockfile && \
+    pnpm build
 
 FROM base AS installer
 
-RUN apk add --no-cache curl bash
+RUN apk update && apk upgrade && \
+    apk add --no-cache curl bash
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Philipinho:
i. Putting them all in one RUN command like this will lead to “ResourceExhausted” error. `pnpm build` alone is resource intensive as it concurrently builds the server, client and packages.

```sh
RUN npm install -g pnpm && \
    pnpm install --frozen-lockfile && \
    pnpm build
```

Ii. There is no such command  as `pnpm cache clean --force` in pnpm. The build will always fail.

I appreciate your willingness to want to contribute, but unfortunately, I will have to close this PR as it doesn’t add much much value to the Dockerfile as is.

PR #894, opened 2025-03-14
feat: auto focus emoji-picker search when opened

GlitchDevX:
Added attribute to make emoji-picker automatically focus the search bar when opened.

## Example

Before:
before

After:
after


Philipinho:

Philipinho:
I unfortunately have to revert this change (https://github.com/docmost/docmost/pull/900).
It scrolls the editor to the top with the emoji picker autofocus enabled. Unless we fix this, we have to revert to the default behavior.

GlitchDevX:
Ah I see. Thanks for the clarification. I'll try to fix this behavior and create another PR if I get it working.

PR #926, opened 2025-03-20
feat: toggle sidebar when users tap outside the sidebar on mobile

savvy-bit:


Philipinho:
Thank you for your work. Looks good. 

We should also close the sidebar on mobile if a page is clicked on. Same for the settings sidebar.

savvy-bit:
I guess it also works for the settings sidebar. The sidebar is closed if a page is clicked on.

PR #937, opened 2025-03-23
[Issue-935] consider a case when there are no members to invite after excluding existing members

savvy-bit:


savvy-bit:
This PR is made based on the [Discussion-935](https://github.com/docmost/docmost/discussions/935)
We need to consider a case when there are no members after filtering existing members.

apps/server/src/core/workspace/services/workspace-invitation.service.ts
```diff
@@ -142,7 +142,11 @@ export class WorkspaceInvitationService {
         }));
 
         // When there is nobody to invite
-        if (invitesToInsert.length < 1) return;
+        if (invitesToInsert.length < 1) {
+          throw new BadRequestException(
+            'No members to invite',
+          );
+        }
```
Philipinho: No need to throw an error.

apps/server/src/core/workspace/services/workspace-invitation.service.ts
```diff
@@ -153,6 +157,9 @@ export class WorkspaceInvitationService {
       });
     } catch (err) {
       this.logger.error(`createInvitation - ${err}`);
+      if (err instanceof BadRequestException) {
+        throw new BadRequestException('There are no members available to invite. Please exclude yourself and those already invited.');
+      }
```
Philipinho, line 161: Same here.

Philipinho (changes requested):

apps/server/src/core/workspace/services/workspace-invitation.service.ts
```diff
@@ -141,10 +141,9 @@ export class WorkspaceInvitationService {
           groupIds: validGroups?.map((group: Partial<Group>) => group.id),
         }));
 
-        // When there is nobody to invite
         if (invitesToInsert.length < 1) {
           throw new BadRequestException(
-            'No members to invite',
+            'There are no members available to invite.',
```
Philipinho, line 146: Let's retain the existing message for now.

PR #993, opened 2025-04-05
Added signup feature

panksrajput:


Philipinho:
Hi, we already have a registration process.
If this mode of registration works for you, you may wish to maintain it in your fork.

Issue #936, opened 2025-03-23
Unable to scroll in Members popup modal


Eruthon:
When opening Space settings and choosing Members tab, I can't scroll down the list of members. Only way is to filter them.
Ability to scroll should be added.

Philipinho:
Thanks for your work.
A fix was added as part of https://github.com/docmost/docmost/pull/1000.

Philipinho:
Fixed https://github.com/docmost/docmost/pull/1000.

PR #996, opened 2025-04-06
fix: add scroll in space settings members table (issue #936)

sanua356:
Fix issue 936. Set scroll to table space settings members.
image


Philipinho:
Thanks for your work.
A fix was added as part of https://github.com/docmost/docmost/pull/1000.

PR #1046, opened 2025-04-22
feat: remember and restore previous route when exiting settings

daoch:
### Summary:
This PR introduces a feature that improves navigation between settings pages. It tracks the user's previous non-settings route and allows them to return to that route after exiting the settings section. 

### Before the change:
- The app doesn't track the previous non-settings page.
- When navigating through various settings pages, the back button consistently returns to the previous settings page instead of exiting the settings entirely.

### After the change:
- When entering a settings route, the last non-settings route is saved.
- After leaving the settings section, the user is returned to the previous non-settings route.
- If there is no previous non-settings route, the user is redirected to `/home`.

### Visuals:
- **Before**: 
before

- **After**: 
after

### Why it matters:
This change provides a smoother, more intuitive user experience, ensuring users return to where they left off before visiting the settings section.

---

### Testing:
- Tested manually by navigating between settings pages and confirming that the "Back" button works as expected.
- Verified that the app defaults to `/home` when no previous non-settings route is found.

---

### Edge Cases:
- Handled case where the user directly lands on a settings page (without visiting a non-settings page first).


Philipinho:
Thanks for contributing.

PR #1058, opened 2025-04-23
Add Persian language

mrbooshehri:
This pull request adds support for the Persian (Farsi) language. As Persian is a right-to-left (RTL) language, additional adjustments may be required to ensure full compatibility with RTL text rendering and layout within the application.


Philipinho:
Translation happens on [Crowdin](https://crowdin.com/project/docmost).
I may consider supporting this language in the future.

Issue #1108, opened 2025-04-28
Hover table style height error causing scrollbar to appear


PolarWS:
Hovering in the middle of the table causes a height error, please see the image

![](https://pan.polarws.moe/f/vznCo/f5b6f0cc2b1382c7cac04e417134dfe.png)

![](https://pan.polarws.moe/f/zRehN/904fab301279530dfdab351d3921e8e.png)

PR #1124, opened 2025-04-30
fix(table-hover): adjust height to prevent unexpected scrollbar

PolarWS:
fix: Hover table style height error causing scrollbar to appear #1108
https://github.com/docmost/docmost/issues/1108

Before

After

The multi-line table has already been tested


Philipinho:

PR #1132, opened 2025-05-02
feat: float image

### LINKED ISSUE #1064: [Feature Request]: Word wrap around images
My group's use case for Docmost involves uploading lots of pictures and it would be great if we could wrap text around left or right aligned pictures.

Here's an example of what I mean: 

Image

I think this should be easy to accomplish by altering the .alignLeft and .alignRight classes in the CSS to have a float attribute and a small margin on the side opposite it is floated.

I can see not everyone preferring this, so ideally an align left/right and float left/right as separate options might be best.

Thanks for all the awesome work on this project!

fuscodev:
- close #1064 

floatimage

### DIFF
--- a/apps/client/src/features/editor/styles/core.css
+++ b/apps/client/src/features/editor/styles/core.css
@@ -171,6 +171,15 @@
   .floatRight {
     float: right;
   }
+  
+  @media (max-width: 575px) {
+    .react-renderer.node-image,
+    .react-renderer.node-image img {
+      float: none !important;
+      width: 100% !important;
+      padding: 0 !important;
+    }
+  }
 }
 
 .ProseMirror-icon {


AndrewSav:
FWIW, in the 5 minutes I've spent testing this, I have not encountered any issues.

AndrewSav:
@Philipinho what needs to happen for this to be merged?

aedelor:
I'm dying for this to make it into an update. :sob:

AndrewSav:
Well we all are. It does not seems that the repo owner is interested in engaging here, though. It's been 4 months since the PR was created, and I asked a question about it above which remained ignored. At this stage, I think it would be wise temper our hopes. For what it's worth, there is also this: <https://github.com/Vito0912/forkmost>

andresbott:
NICE!!! i want this

darkshenzi:
When will this feature be available? @Philipinho

Bettelmage:
Please, please, please add this feature!

PR #1153, opened 2025-05-11
feat: create new page from mention

### LINKED ISSUE #1010: [Feature Request] - Create new page through Mention
Many Markdown editors allow for new pages to be created directly from the editor

It would be amazing if the "Mention" component included the page creation feature if no page was found by the search. (illustrated below)

The page could be created as a child of the one currently being edited and automatically linked to the mention that created it.

Image

Thank you for the amazing software !

fuscodev:
- close #1010 

newpage


Philipinho:
I noticed pages created from sub-pages get created in the root. Ideally, I think they should be in their relative parent root.
What do you think?

fuscodev:
> I noticed pages created from sub-pages get created in the root. Ideally, I think they should be in their relative parent root. What do you think?

Yes, I fixed it

Philipinho:
Thank you. Merged 🚀

PR #1184, opened 2025-05-15
fix: sanitize tree filename in space export

### LINKED ISSUE #837: Forward slash / in page title creates folder in export zip
If you have forward slashes in your page titles, when you export the space, this creates folders within the export.

For example, if your page is named '**Birthday Party 12/5/25**' then the export will save the page as **25.md** within the folder structure Birthday Party 12 → 5 → 25.md

Image

fuscodev:
- fix #837


Philipinho:
This is now fixed.

PR #1195, opened 2025-05-18
feat: space graph view

fuscodev:
graphview

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -45,6 +45,7 @@
     "react-dom": "^18.3.1",
     "react-drawio": "^1.0.1",
     "react-error-boundary": "^4.1.2",
+    "react-force-graph-2d": "^1.27.1",
     "react-helmet-async": "^2.0.5",
     "react-i18next": "^15.0.1",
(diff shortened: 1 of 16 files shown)


Vito0912:
Worked really well, even for larger documents. But one "issue" (mor of UX) I noticed is that they become very crowded very quickly, if you have connections between shared nodes.

As a suggestion: Maybe make the nodes larger with more connections to stand out (and increase distance) and add transparency to the edges?

Before (as you can see many are on one place etc.)
grafik
After some tweaking (Still not perfectly):
grafik

fuscodev:
Do you want to share your changes?

Vito0912:
It still has a few bugs and is not perfect yet. 
I also want to find a way to remove/reduce the initial moving time, but this is what I got after a full "rewrite"/refactoring of the. I basically rewrote everything to better understand how it works. 
So, the diff is a bit messy: https://github.com/Vito0912/forkmost/commit/66460a1cf9f9782352d1f1ec6a878aada344623f

Philipinho:
Thanks for your contribution, but no plans to support this at this time.

PR #1215, opened 2025-05-28
feat: use environment variable in docker-compose.yml for configurability

tuladhar:
This PR updates the docker-compose.yml file to replace hard-coded configuration values with environment variable substitutions. Instead of directly editing the docker-compose.yml file to change configurations during [the installation](https://docmost.com/docs/installation#replace-the-default-configs), users can now supply values via a .env file.

> [!IMPORTANT]
> If this gets merged, we need to merge this PR in the documentation: https://github.com/docmost/docs/pull/14

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,9 +7,9 @@ services:
       - db
       - redis
     environment:
-      APP_URL: 'http://localhost:3000'
-      APP_SECRET: 'REPLACE_WITH_LONG_SECRET'
-      DATABASE_URL: 'postgresql://docmost:STRONG_DB_PASSWORD@db:5432/docmost?schema=public'
+      APP_URL: "$APP_URL"
+      APP_SECRET: "$APP_SECRET"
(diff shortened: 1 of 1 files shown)


ctmbl:
In my (totally personal and humble) opinion, this kind of change is not desirable in general.

When you want to deploy to test a new app it's pretty cool if the provided docker-compose works out-of-the-box, because you're looking for a fast try-and-test configuration.

Yes the provided docker-compose is not configurable, and yes it is **really not secured and shouldn't be used as-is in production**. But replacing default value by some environment variable stored in a `.env` is a [Docker feature and workflow](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file), and thus not really the scope of Docmost documentation (once again in my opinion).  
Maybe the docs should instead precise that this is **not a production-ready compose file** and point to the appropriate documentation.

Philipinho:
I prefer to leave this to the user.

PR #1229, opened 2025-06-07
security fix: prevent access to attachment link after page is unshared

### LINKED ISSUE #1228: Attachment download links remain accessible after public sharing is disabled
**Affected Version:** v0.20.4

## Description

There is a security-related issue in how Docmost handles public file attachments. When a page is publicly shared and a visitor clicks the **Download** button on an attachment, a direct file URL is generated. This file remains accessible even after the public share link is revoked.

## Problem Summary

- While the page is shared via a public link (e.g. `https://docs.example.com/share/0ysbhetox0/p/<name-of-file>-qx22bnGfPs`), a visitor clicks the download button for an attached file.
- This triggers generation of a direct download link like:  
  `https://docs.example.com/api/files/public/01974a08-2a63-7069-aee2-64fa52501b05/filepdf?jwt=<token>`
- Afterwards, the public share is disabled.
- Result:
  - The shared page URL is no longer accessible (expected).
  - The direct file URL remains publicly accessible without authentication (unexpected).

This issue only occurs **after the download button was clicked** while the page was public. The generated file link remains valid regardless of the current page visibility.

## Expected Behavior

When public sharing of a page is disabled:

- The page should no longer be accessible without authentication.
- Any previously exposed file download links should also become invalid or require login.

## Security Impact

- File attachments remain accessible to third parties, even after the page is no longer public.
- This may lead to unintended data exposure.
- Users may incorrectly assume that removing a share link also revokes all public access to associated files.

## Recommendation

Ensure that file download links generated in the public view are only valid while the parent page is shared. Once the share is disabled, file access should be restricted accordingly — either by invalidating the link or requiring authentication.

fuscodev:
- fix #1228 

Check if the page is still shared before serving attachments, without waiting for JWT expiration.

### DIFF
--- a/apps/server/src/core/attachment/attachment.controller.ts
+++ b/apps/server/src/core/attachment/attachment.controller.ts
@@ -50,6 +50,8 @@ import { validate as isValidUUID } from 'uuid';
 import { EnvironmentService } from '../../integrations/environment/environment.service';
 import { TokenService } from '../auth/services/token.service';
 import { JwtAttachmentPayload, JwtType } from '../auth/dto/jwt-payload';
+import { ShareService } from '../share/share.service';
+import { ShareInfoDto } from '../share/dto/share.dto';
 
 @Controller()
 export class AttachmentController {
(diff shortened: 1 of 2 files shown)


Philipinho:
I commented on the issue https://github.com/docmost/docmost/issues/1228#issuecomment-2952777123.

PR #1256, opened 2025-06-13
feat: add extra ligatures

Webblitchy:
I created new typographic conversions (ligatures)
- There is already a lot of existing conversions : ->, !=, +/-
(full list : https://tiptap.dev/docs/editor/extensions/functionality/typography#rules)
        
- I added 3 missing conversions : =>, >=, <=

Here is a gif demonstrating the changes:

extra-ligatures


Philipinho:
Will consider in the future if there is need for it.

PR #1272, opened 2025-06-19
feat: Ctrl/Cmd+S: prevent 'Save As' dialog

### LINKED ISSUE #1263: Ctrl+S: prevent 'Save As' dialog
I know it saves automatically, but it would be great if the browser's usual "Save As" dialog didn’t pop up. I keep triggering it out of muscle memory because I'm used to hitting save every now and then.

fuscodev:
- close #1263

### DIFF
--- a/apps/client/src/features/editor/page-editor.tsx
+++ b/apps/client/src/features/editor/page-editor.tsx
@@ -150,11 +150,6 @@ export default function PageEditor({
     ];
   }, [ydoc, pageId, remoteProvider, currentUser?.user]);
 
-  const debouncedSendSaveCommand = useDebouncedCallback(() => {
-    const payload = 'forceSave';
-    remoteProvider.sendStateless(payload);
-  }, 300);
-
   const editor = useEditor(
     {
       extensions,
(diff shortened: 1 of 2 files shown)


apps/client/src/features/editor/page-editor.tsx
```diff
@@ -161,6 +166,11 @@ export default function PageEditor({
         scrollMargin: 80,
         handleDOMEvents: {
           keydown: (_view, event) => {
+            if ((event.ctrlKey || event.metaKey) && event.key === 's') {
+              event.preventDefault();
+              debouncedSendSaveCommand();
```
Philipinho: We won't be needing this. I.e., the `debouncedSendSaveCommand` function.

apps/client/src/features/editor/page-editor.tsx
```diff
@@ -161,6 +161,10 @@ export default function PageEditor({
         scrollMargin: 80,
         handleDOMEvents: {
           keydown: (_view, event) => {
+            if ((event.ctrlKey || event.metaKey) && event.key === 's') {
```
sanua356: Hi! Your solution will work, but it is not very reliable. When switching keyboard layouts, this function will stop working (Or if CapsLock button pressed). It is better to use event.code, which does not depend on the user's chosen language.

**Example Problem**
image

**Suggestion**
```suggestion
            if ((event.ctrlKey || event.metaKey) && event.code === 'KeyS') {
```

fuscodev: Thanks, I'll fix soon

PR #1292, opened 2025-06-24
Allow admins to change the password of members

Vito0912:
I understand that this feature might not be possible for enterprise use. If that's the case, please close this PR. 

I opened it in case it could be added. It does not let admins change the owner's email. Everything else can already be done by an admin (deleting an account, creating a new one) or by making a new account with the admin role.

It also lets the person know who changed the password.


Philipinho:
No plans to support this.
The password reset functionality is there for this purpose.

Issue #22, opened 2024-06-30
SSO via OIDC?


osresearch:
Is single sign on support on the roadmap? Being able to create users and manage permissions via OIDC is very useful not only for enterprise applications, but also for smaller communities.  On the https://v.st domain we have mastodon, matrix, gitea, nextcloud, hedgedoc, etc all centrally managed via keycloak so that onboarding a new member only requires adding their account in a single place.

will2hew:
@Philipinho would you accept PRs to add this functionality

archef2000:
Would also like to help.

will2hew:
@Philipinho not to worry! I have done some exploratory work in a fork. 

My suggested changes:

1. The initial scope would include a per-workspace OIDC configuration (only one per tenant), which would also include a list of domains that will have accounts JIT created upon login. It will also add a configuration option for enforcing OIDC. 
2. This configuration will live in a seperate column to the existing `settings` column, as we don't want it getting selected by default and including the client secret to the frontend. 
3. I am also proposing some slight simplifications to token management, I don't believe we're actually using the refresh token at all so I'd just suggest that we use a server-side HTTP-only cookie to hold the JWT. 
4. We'd use the `passport-openidconnect` strategy. I haven't fully scoped this out, but I think we'd use a `passport-custom` strategy as a wrapper to then create a new OIDC strategy at runtime - to support per-tenant OIDC configuration.

Let me know what you think!

will2hew:
@Philipinho in reference to #35 where I would second that neither attachments or profile pictures should be public, we can actually embed the permissions into the JWT - this is one of their main advantages - and avoid a database call for permissions check. This is secure as we will sign it using our secret, and can verify we were the one that signed it. So if a client provides a valid JWT saying it contains the `view:attachment` permission for example, without a database call we can verify it.

I'd propose making permissions embedded in JWTs part of this PR

Philipinho:
@will2hew I have yet to give a thought to your OIDC process.

Talking about the attachment permissions, do you mean putting the permission inside the sign-in JWT token? or or a separate time-based JWT generated for each attachment?

archef2000:
The problem is that a group can be removed or added to a user after the JWT is issued.

will2hew:
@Philipinho correct, I would suggest putting permissions inside the JWT. 

[...]

PR #1311, opened 2025-06-29
Add optional password protection to shared pages

### LINKED ISSUE #1098: [Feature] Use a password when sharing a page
The ability to share pages has been released, which is very exciting!🎊🎊🎊

I believe it would be even better if there was an option to use a password when sharing.❤️❤️❤️

### LINKED ISSUE #1308: feat request
create password for public page like allow setup password when publishing

Vito0912:
This adds password protection to the shared pages. 

This fixes #1098 and it's duplicate fixes #1308

Only users with full access should be able to remove a password to make sure members cannot take it off sensitive files. They can disable the share and enable it again at any time, but this will create a new link.

The password is stored securly in the DB. Thus it cannot be displayed after entering

grafik
grafik
grafik

### DIFF
--- a/apps/client/src/features/share/components/share-modal.tsx
+++ b/apps/client/src/features/share/components/share-modal.tsx
@@ -9,14 +9,18 @@ import {
   Text,
   TextInput,
   Tooltip,
+  PasswordInput,
+  Divider,
 } from "@mantine/core";
-import { IconExternalLink, IconWorld } from "@tabler/icons-react";
+import { IconExternalLink, IconWorld, IconLock, IconLockOpen } from "@tabler/icons-react";
 import React, { useEffect, useMemo, useState } from "react";
 import {
   useCreateShareMutation,
   useDeleteShareMutation,
   useShareForPageQuery,
   useUpdateShareMutation,
+  useSetSharePasswordMutation,
+  useRemoveSharePasswordMutation,
 } from "@/features/share/queries/share-query.ts";
 import { Link, useParams } from "react-router-dom";
 import { extractPageSlugId, getPageIcon } from "@/lib";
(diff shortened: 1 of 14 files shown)


Philipinho:
However, we plan to include controls like this in the enterprise edition.

Vito0912:
I was sadly already afraid that something like this might happen. 
It would be helpful if there were a list where you could see such plans/features (or maybe there is one already?). 

For others so that this doesn’t get lost and for ones who find this PR: I will always keep [Forkmost ](https://github.com/Vito0912/forkmost) up to date as long as I use ~~Docmost~~ Forkmost myself and will keep adding my features. No paywall ever.

I understand that money needs to be made in some way, but I do not understand why, like the SSO features, PRs are not being merged just because they might be added to an enterprise plan in the future (TM), even though they could be altered to fit these needs e.g. for the cloud version. Still I acknowledge your major work on that project without that all wouldn't exists, so defintily a thank you for this

Then please close the PR if this is not wanted

Philipinho:
I appreciate your eagerness to want to contribute.
However, I have addressed this on the SSO issue in the past.
And I quote below:
> I believe it is important to create a balance to have an avenue to help fund the continued development of the project.
Source: https://github.com/docmost/docmost/issues/22#issuecomment-2521407735

Without an avenue to generate revenue, It simply will not be possible to maintain the project for a very long time.

I would not want to be in a position where I would be considering moving features from the OSS edition to enterprise. That is why I have to be upfront about it.

To you, SSO in enterprise is fine, but to many as evidenced on the issue, it is not.
While my ideas for the project might not be what you want, it is not possible to please everyone.

Being open-source doesn’t mean we have to accept every idea, pull request or feature implementation.

This also applies to https://github.com/docmost/docmost/pull/1311.

XJY00:
#1719

PR #1319, opened 2025-07-01
Added 2FA: TOTP

### LINKED ISSUE #819: [Feature Request] 2FA verification for user login
Is there a plan to add 2FA verification function for user login?
happy to see that

### LINKED ISSUE #1315: Feature-Request: 2 Factor Authentication
Hi, 
Did I missed it or is there no 2-Factor Authentication? I would like to use it.
Would like to use TOTP with QR and / or Yubikey.
Thanks in advance.

Vito0912:
This was decided to be in the enterprise version. See https://github.com/docmost/docmost/issues/1315#issuecomment-3021699265

I am opening this PR only to make this visible. 

SSO/OICD is, unfortunately, often put in a higher tier - So while I also don't think that is correct in an "Open-Source" world, it's not unfair or anything. 
However, putting basic and necessary 2FA like TOTP behind an enterprise tier (not even the basic cloud version) is just a security risk. 

This PR is not the best version of such implementation and I might rewrite it in the future to be more dynamically, but it should work for now. 
I already felt that even simple necessary security features are being paywalled, why I wanted to make a general PoC first (and because I will use it for myself the time is not wasted). Sadly now even other, already implemented features, are hidden behind unclear and intransparent pricing (now the prices for the enterprise tier are gone, so you have to contact them directly). - At the time of writing this PR there are even plans mentioned not even existing currently on the website, so it is clear that they want to expand the paid offer, which is fine, but as it looks like at the cost of security for everyone else

#- Other companies and products require users, even on free plans, to enable 2FA before they can use the service.

Please close this PR if this really should become part only of the ee version, so I can at least say I tried. I am happy to rewrite it to fit more generally if wanted.

## Techincal

Everything should be saved securely in the database and can only be read with the app secret. If the app secret is changed, it will no longer work. Backup codes are stored as a hash using a one-way function. The time windows is +-1 which allows for human slowness ;) or slightly off sync clocks.

Fixes #150 
Fixes #819 
Fixes #1315


Philipinho:
I appreciate your eagerness to want to contribute.
However, I have addressed this on the SSO issue in the past.
And I quote below:
> I believe it is important to create a balance to have an avenue to help fund the continued development of the project.
Source: https://github.com/docmost/docmost/issues/22#issuecomment-2521407735

Without an avenue to generate revenue, It simply will not be possible to maintain the project for a very long time.

I would not want to be in a position where I would be considering moving features from the OSS edition to enterprise. That is why I have to be upfront about feature development. And that is why we have both the Open-Source and Enterprise editions.

While my ideas for the project might not be what you want, it is not possible to please everyone.

Being open-source doesn’t mean we have to accept every idea, pull request or feature implementation.
Rather, It gives you, the user, the freedom to maintain your own fork and implementation.

I understand this may not be what you wish to hear, but we have to make the product viable so it benefits everyone in the long run.

This also applies to https://github.com/docmost/docmost/pull/1311.

Vito0912:
> Without an avenue to generate revenue, It simply will not be possible to maintain the project for a very long time.

I completely agree with you. No questions about that. However, prices have become less clear since started offering multiple subscription models in the backend. 
Yes, the standard plan got cheaper, but only if you have at least three users. For two users, it actually became more expensive.
Enterprise (which is the important for self-hosting) plan now is completely intransparent.

But here is my main concern: SSO, (planed) page passwords, and now even (planned) 2FA are only available for enterprise customers. So only enterprise users can access these features. This seems to be your target group then. - Still all this are features "normal" customers might want - e.g. "Remove branding in public pages", "Email support" (and maybe even "Confluence import") are things I fully understand and are excluded from what I mean
Why can’t these features be available for individual users, while still making enterprises pay? 
It would be possible. You could add an extra license for these features, where the user has to confirm it is for educational or personal use or whatever. 
These personal customers will never pay, especially since they now need to get an quote just to start. Or limit it to x users.

> I would not want to be in a position where I would be considering moving features from the OSS edition to enterprise.

Luckily, that is not even possible (TM) because of the license you use, which is a good thing. (And just to be clear, there have to be EE modules, I not say, open source everything). 
But to pick up the thought above, they could be implemented via that way (Just be usable if signed a agreement)

> While my ideas for the project might not be what you want, it is not possible to please everyone.

I completely understand that. For example: https://github.com/docmost/docmost/pull/1292  
I fully understand why this was closed, and I know it is not possible to make software that works for everyone. But I also never said so. I mean all things that I made a PR to where not closed because "it does not works for everyone", but "enterprise"
It’s just that you put a basic 2FA feature in the **enterprise** tier. I litteraly know no company that does not support 2FA. And very surely no company that offers no 2FA if you are a paying customer - standard
\* From companies who offer 2FA at any place

> Being open-source doesn’t mean we have to accept every idea, pull request or feature implementation.
Rather, It gives you, the user, the freedom to maintain your own fork and implementation.

Did not say that, just to be clear. Only criticized not even considering pull requests for basic security features and putting them in a tier that is not accessible to a normal user.


## But

If you have no problem with it, can you provide a list with features you plan to publish under the enterprise plan?
It would really help if it's even worth to create a PR

XJY00:
#1719

Issue #1296, opened 2025-06-25
Custom emoji for Callouts


Blankeos:
Currently there are only 4 types of callouts. Would be cool if we could use a custom emoji for it like in Notion.

<img width="841" alt="Image" src="https://github.com/user-attachments/assets/52dc05af-321d-4000-b313-a1950559c597" />

Notion

<img width="957" alt="Image" src="https://github.com/user-attachments/assets/3855d320-236b-4731-8192-bf90b63823de" />

Feel like integrating the existing emoji picker here would be good enough!

Vito0912:
I added this in a commit in a fork: https://github.com/Vito0912/forkmost/commit/f366d0fb353f72ca87e3cd9a55bb54d10bf97783. However, I can't open a PR because Git will not let me cherry-pick to a branch that is in sync with docmost. This is due to some issues with the submodules being used (I'm not sure why as I don't use submodules with git).
As covered by the license this can be used as a PR

![Image](https://github.com/user-attachments/assets/b5039a39-22c4-4c46-8805-77e300330fe9)

PR #1323, opened 2025-07-02
Add custom emoji to callout

Vito0912:
See #1296 for example


Philipinho:

PR #1336, opened 2025-07-04
fix ctrl-a for codeblocks

Vito0912:
related to an issue I can't find anymore.  

Makes it so ctrl-a only selects content in a code block, which makes working with code blocks much easier, especially if transferring a mermaid diagram or something similar.


Philipinho:
Thanks for your contribution.

PR #1381, opened 2025-07-18
feat(EE): MFA implementation

Philipinho:
- Add TOTP-based two-factor authentication
- Add backup codes support
- Add MFA enforcement at workspace level
- Add MFA setup and challenge UI pages
- Support MFA for login and password reset flows
- Add MFA validation for secure pages


compgeniuses:
I am sorry to bring this Up, but we should never treat Security as Entepirse, everyone deserves security, yes already SCIM, SSO and others ar ein enteprirse, but for 2FA, this should definitely also be available to All versions.. Jus tmy thinking

Issue #1259, opened 2025-06-16
Improve Mermaid chart theme compatibility with dark mode


WebKenth:
When using the application in dark mode, the default theme for Mermaid charts lacks sufficient contrast and readability. Enhancing the Mermaid chart styling for dark mode would improve user experience and visual consistency.

Suggested improvement:

Adjust Mermaid chart theme colors to ensure better visibility in dark mode

Consider using Mermaid’s built-in dark themes or customizing CSS variables accordingly

![Image](https://github.com/user-attachments/assets/0dfceb83-0723-4a5b-8da3-d9bc7ca1b969)

PR #1397, opened 2025-07-22
feat: set mermaid theme based on computed color scheme

aschaber1:
Use Mantine's `useComputedColorScheme` hook to dynamically configure mermaid's theme.
- When the computed color scheme is "light", the theme is set to "default".
- Otherwise, it is set to "dark".

List of [available themes](https://mermaid.js.org/config/theming.html#available-themes):
- `default` - This is the default theme for all diagrams.
- `neutral` - This theme is great for black and white documents that will be printed.
- `dark` - This theme goes well with dark-colored elements or dark-mode.
- `forest` - This theme contains shades of green.
- `base` - This is the only theme that can be modified. Use this theme as the base for customizations.

Closes: #1259


Philipinho:

PR #1404, opened 2025-07-24
Update main.ts

Rogermax:
This log is not correct... in reality is listening on 0.0.0.0 I thought checking the logs that this was the problem to not connect from an nginx... But checking the code seems in reality listen well in 0.0.0.0.... so let's put the log correct!


Philipinho:
In most cases, 127.0.0.1 works.

PR #1438, opened 2025-08-01
fix: set mermaid theme based on computed color scheme

aschaber1:
Hello,

after #1397 failed and you had to revert in #1412 - sorry for that. 

I now come with a second try, this time I have successfully tested it locally and it worked. Tried all transitions:
- light -> dark
- light -> auto
- dark -> light
- dark -> auto
- auto -> light
- auto -> dark

It works.
Note: There seems to be a slight (we're talking low low ms range) delay between docmost changing theme and mermaid, that is sometimes noticeable. I have tried it with Firefox.


Philipinho:

Issue #1194, opened 2025-05-17
Bug: Picture on publicly shared note not showing


ayoayco:
Hello 👋 Thanks for working on Docmost! Want to report this behavior which might be unexpected?

I have a [publicly shared note](https://notes.ayo.run/share/qgihgf2kym/p/media-NEJmxq3bsP) with an attached picture, but it is getting `401 Unauthorized`: https://notes.ayo.run/api/files/0196dfbc-c639-7798-99af-fde85ed67f8b/2024_5846.JPG

Philipinho:
This usually happens if the attachment was copied and pasted from another page.

ayoayco:
@Philipinho thanks I confirmed that is the case. Added a picture to the note directly and that is working fine.

PR #1449, opened 2025-08-05
Support cross-page attachment sharing with sharedPageId

need4swede:
Adds support for sharing attachments across pages by introducing an optional sharedPageId in the JWT payload. Updates attachment controller validation logic to handle cross-page access, modifies token generation to include sharedPageId, and updates the share service to set this field when generating tokens for shared attachments.

This should fix the issue where copy/pasted images don't work on public (shared) pages.
See issue #1194 

## Summary by CodeRabbit

* **New Features**
  * Added support for accessing attachments shared across different pages, with strict checks to ensure workspace and space consistency.
* **Improvements**
  * Enhanced attachment sharing tokens to include optional cross-page access information for more flexible sharing scenarios.

### DIFF
--- a/apps/server/src/core/attachment/attachment.controller.ts
+++ b/apps/server/src/core/attachment/attachment.controller.ts
@@ -232,10 +232,36 @@ export class AttachmentController {
       !attachment ||
       attachment.workspaceId !== workspace.id ||
       !attachment.pageId ||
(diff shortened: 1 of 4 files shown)


Philipinho:
I don't think this would make sense if we eventually support page-level permissions, which was the basis of this design decision.

PR #1467, opened 2025-08-13
feat: table row/column drag and drop

Saul-Mirone:
Add drag and drop feature for table rows and columns.

https://github.com/user-attachments/assets/341cf21b-3eaf-4877-8e06-7b0a836f2148

## Summary by CodeRabbit

* **New Features**
  * Full table drag-and-drop: draggable row/column handles, live preview, drop indicators, auto-scroll during drag, and move-on-drop with selection preservation.

* **Style**
  * UI styles for rotated horizontal handle, drag previews and drop indicators; hides default drop cursor while dragging; added editor container hook.

* **Chores**
  * Added positioning helper dependency to improve handle/preview/drop-indicator placement.


Philipinho:
Works really well.

Is it possible to disable the default editor indicator when a table drag is ongoing? 

https://github.com/user-attachments/assets/76b1cb45-57a0-4461-aa96-6cdcc3d81c87

Saul-Mirone:
@Philipinho I've added this optimization from https://github.com/docmost/docmost/pull/1467/commits/4954d7f2fa8db552d92c559af4e5f4a618ae873b.

Philipinho:
Larger tables have scroll areas.
I am unable to scroll with my trackpad while dragging in such cases. 
We should support scrolling vertically and horizontally with the trackpad while dragging. 
What do you think?

In this short demo, I can scroll without dragging but can't while dragging if I want to place the dragged column at the end of the table.

https://github.com/user-attachments/assets/28f168b4-cd06-4fe8-b1f2-f124be801eba

Saul-Mirone:
@Philipinho Hi, I've added the auto scroll feature.

Philipinho:
My observations.
Safari seem to allow trackpad scrolling on drag by default.
Chrome seem not to, even with the recent change, it doesn't work.

Then on Firefox, reordering does not work at all.
Would love your thoughts on this and if you can reproduce.

Firefox video:

https://github.com/user-attachments/assets/a65d5033-478a-4f60-a2a7-836e83401538

Saul-Mirone:
@Philipinho I've fixed the issue in firefox. The drag and drop API in firefox has some bugs. Chrome autoscroll works on my machine:

https://github.com/user-attachments/assets/54e5a6df-638b-4ef6-a098-c6b1e9ee8730

mustafa-yilmaz:
That feature should only work for users with edit access on the page.

If it will also work for users with only view access, changes (alterations) that those users will make should not effect what other users see on the page.

PR #1478, opened 2025-08-19
Fixing extra page bug on print

### LINKED ISSUE #1464: Print PDF adds extra blank page to end of job
When selecting "Print PDF", a blank page is always appended to the end of each document.

Currently I need to select a specific range of pages to print to ensure the extra page isn't included in the print job.

Any help on this would be greatly appreciated.

Eshwar1212-maker:
This issue fixes #1464

The extra blank page was caused by Mantine’s <AppShell.Main> applying min-height: 100vh in print mode. In Chrome, 100vh maps to the full page box (~1056px) while the printable area is slightly smaller (~982px), which forced an overflow onto a second empty page.

 I added min-height: auto to the existing class in 'print.css, so the container shrink-wraps to its content. This removes the phantom page without affecting normal screen layout.

https://github.com/user-attachments/assets/61b0798a-51f2-4028-94f2-d8107389f9ff

### DIFF
--- a/apps/client/src/features/editor/styles/print.css
+++ b/apps/client/src/features/editor/styles/print.css
@@ -7,5 +7,6 @@
 
     .mantine-AppShell-main {
         padding-top: 0 !important;
+        min-height: auto !important;
     }
 }


Eshwar1212-maker:
@Philipinho Let me know what you think!

apps/client/src/features/editor/styles/print.css
```diff
@@ -7,5 +7,6 @@
 
     .mantine-AppShell-main {
         padding-top: 0 !important;
+        min-height: auto !important;
```
Philipinho, line 10: Let’s put this in CSS print.

Eshwar1212-maker: Great catch! Ill make that adjustment later today.

Eshwar1212-maker: @Philipinho Hey so my change was in 'print.css' I cant find any other file called print css. 

I noticed there’s a core.css file that already has some global @media print rules. Did you mean I should move this override there instead of the editor’s print.css? Just want to confirm before I adjust it.

Philipinho: Oh, I see its actually in the right place `@media print`.
I did not look at the full file content the first time.
Thank you.

Eshwar1212-maker:
@Philipinho  hey, I was just wondering if there's anything else I need to do for a merge? Thank you.

Philipinho:
I appreciate it.

Eshwar1212-maker:
Sorry for rushing you lol, thank you so much! @Philipinho

PR #1562, opened 2025-09-15
fix: handle empty replace term in search and replace functionality

### LINKED ISSUE #1537: [BUG] Unable to replace with empty string
When replacing text with the empty string, no change is made and an error is produced in the console stating ```Uncaught RangeError: Empty text nodes are not allowed```

**Steps to Reproduce:**
1. Press `ctrl+f` to open the search bar;
2. Type any text to search for;
3. Click the replace icon;
4. Click the `Replace` or `Replace all` button without entering any replace text;

**Expected Behavior:**
The matches found by the search function should be eliminated when replaced by the empty string.

**Actual Behavior:**
Nothing gets replaced.

**Error Message:**
```Uncaught RangeError: Empty text nodes are not allowed```

**Environment:**
• Docmost Version/Build: Self-hosted/v0.22.2

**Possible Cause:**
The [replace](https://github.com/docmost/docmost/blob/091e790b83edd13435267dc5cff84c50b5440fc9/packages/editor-ext/src/lib/search-and-replace/search-and-replace.ts#L173) and [replaceAll](https://github.com/docmost/docmost/blob/091e790b83edd13435267dc5cff84c50b5440fc9/packages/editor-ext/src/lib/search-and-replace/search-and-replace.ts#L206) methods do not support an empty string as replaceTerm.

**Possible Solution:**
Only create marksSet and insert them if replaceTerm is not the empty string
```
// Only insert new text if replaceTerm is not empty
// ProseMirror doesn't allow empty text nodes
if (replaceTerm !== "") {
```

Pleasurecruise:
Fix #1537 

- Fix 'Empty text nodes are not allowed' error when replace field is empty
- Allow empty replace term to perform batch deletion of matched text
- Update both replace() and replaceAll() functions to check for empty replaceTerm

### DIFF
--- a/packages/editor-ext/src/lib/search-and-replace/search-and-replace.ts
+++ b/packages/editor-ext/src/lib/search-and-replace/search-and-replace.ts
@@ -195,9 +195,13 @@ const replace = (
     
     const marks = Array.from(marksSet);
     
-    // Delete the old text and insert new text with preserved marks
+    // Delete the old text
     tr.delete(from, to);
-    tr.insert(from, state.schema.text(replaceTerm, marks));
+
+    // Only insert new text if replaceTerm is not empty (allows for deletion when replaceTerm is empty)
+    if (replaceTerm) {
+      tr.insert(from, state.schema.text(replaceTerm, marks));
+    }
     
     dispatch(tr);
   }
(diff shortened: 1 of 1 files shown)


Philipinho:
Thanks for your contribution.

PR #1563, opened 2025-09-15
fix: sanitize page titles in export utils

### LINKED ISSUE #1552: Export: Filenames containing a slash create unintended subdirectories
**Description:**
When a note is titled **Foo 12/2020**, the exported file is saved as **Foo 12/2020.md**.
Because / is interpreted as a directory separator by the filesystem, this results in a subfolder **Foo 12/** containing a file named **2020.md** instead of a single file.

**Expected behavior:**
Invalid or reserved characters in filenames (e.g. /) should be sanitized and replaced with safe alternatives (e.g. - or _).
In this case, the exported file should be named **Foo 12-2020.md** or even **Foo_12-2020.md**, if the space is also sanitized.

Pleasurecruise:
Fix #1552 

Replaces raw page titles with sanitized versions using the sanitizeFileName helper in getPageTitle to ensure safe filenames during export.

## Summary by CodeRabbit

* **Bug Fixes**
  * Exported file names are now automatically sanitized to remove or replace invalid characters, improving compatibility across operating systems and file storage services.
  * Reduces failures when downloading, syncing, or sharing exports due to problematic characters in titles.
  * Exports for pages without titles continue to use “untitled,” ensuring consistent behavior.

### DIFF
--- a/apps/server/src/integrations/export/utils.ts
+++ b/apps/server/src/integrations/export/utils.ts
@@ -5,6 +5,7 @@ import { validate as isValidUUID } from 'uuid';
 import * as path from 'path';
 import { Page } from '@docmost/db/types/entity.types';
 import { isAttachmentNode } from '../../common/helpers/prosemirror/utils';
+import { sanitizeFileName } from '../../common/helpers';
 
 export type PageExportTree = Record<string, Page[]>;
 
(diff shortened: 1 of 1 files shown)


Philipinho:
This is now fixed.

PR #1601, opened 2025-09-26
fix: fix .env file not loaded in docmost container

### LINKED ISSUE #1566: Unable to send email — no error in production logs
Hi everyone,

I’m using Docmost version 0.23.1, and we’ve noticed that emails aren’t being sent to production without any logs being generated.

Before opening this issue, I reviewed [discussion 744](https://github.com/docmost/docmost/discussions/744), which links to [issue 205](https://github.com/docmost/docmost/issues/205) and [issue comment 2309861444]. However, none of these discussions provided a solution to my problem.

Here’s my anonymized configuration:

```
MAIL_FROM_ADDRESS=noreply@xxx.com
MAIL_FROM_NAME=Docmost
SMTP_HOST=smtp.gmail.com
SMTP_PASSWORD=xxx
SMTP_PORT=587
SMTP_SECURE=false
SMTP_USERNAME=noreply@xxx.com
```

I want to clarify that the password is an application password, and it worked until today. I’m confident that these credentials are valid because I’ve tested them with another app (not Docmost).

The most frustrating aspect is that there are no logs indicating any issues.

If anyone can help me resolve this problem or has experienced a similar issue, I’d greatly appreciate any assistance.

Thank you.

EvanSchleret:
I'm using Coolify to use docmost and environment variables are not loaded into docmost container (all other env variables that aren't in the docker-compose file, including mail configuration). This PR just adds the loading of the `.env` file contained in the `docker-compose.yml` directory.

Close #1566

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
       APP_SECRET: 'REPLACE_WITH_LONG_SECRET'
       DATABASE_URL: 'postgresql://docmost:STRONG_DB_PASSWORD@db:5432/docmost?schema=public'
       REDIS_URL: 'redis://redis:6379'
+    env_file: ".env"
     ports:
       - "3000:3000"
     restart: unless-stopped


EvanSchleret:
Actually, it’s not a proper Docmost issue with the compose file because it’s loaded from the coolify template directory. However, I still think it’s a good idea to allow loading of environment variables from .env.

https://github.com/coollabsio/coolify/pull/6705

Issue #1646, opened 2025-09-29
[Feature Request] - Set HOME page for a space


millab:
On mobile devices in particular, it is hard for users to navigate the space with the sidebar hidden. They are greeted with only a list of recently updated pages. 

Can we please specify a home page for each space, which is shown instead of the current overview? 
This is something that works great in Confluence

For example: 
I would like to create a page within the space, and set it to be the default page you see when first opening the space. It should replace the 'Overview' option in the sidebar too.

arjitc:
This would be pretty nice, like a pinned/starred page which is the primary page to show when a space/category is selected, where things like say a FAQ/Information about the space can be setup.

mustafa-yilmaz:
Also requested at #1616 .

millab:
> Also requested at [#1616](https://github.com/docmost/docmost/discussions/1616) .

Actually that was about setting a default space - this is about having a home page within a space

mustafa-yilmaz:
> > Also requested at [#1616](https://github.com/docmost/docmost/discussions/1616) .
> 
> Actually that was about setting a default space - this is about having a home page within a space

My fault, sorry.

Eshwar1212-maker:
@millab @arjitc Hey guys, I just fix this if you guys can assign the issue to me? It's about 95% on I'll upload a video tomorrow.

Let me know if you guys think it works if this is what you want? 

I kept the overview page however, if I use your set page to be a homepage, it replaces the overview page every single time to navigate to their homepage instead of the overview page, and if a user does not have a homepage selected, they can simply have the overview page there.

I don't think it makes sense to fully replace the overview page because many users by default might not set a homepage. But the same functionality is practically there aware if you set a homepage and you go back to that space it always directs you immediately to that homepage not the overview page.

So for every single page, I just added the option to make that specific page the homepage(only one homepage can be there obviously at a time). And a small icon that shows that that's the homepage.

millab:
> [@millab](https://github.com/millab) [@arjitc](https://github.com/arjitc) Hey guys, I just fix this if you guys can assign the issue to me? It's about 95% on I'll upload a video tomorrow.
> 
> Let me know if you guys think it works if this is what you want?
> 
> I kept the overview page however, if I use your set page to be a homepage, it replaces the overview page every single time to navigate to their homepage instead of the overview page, and if a user does not have a homepage selected, they can simply have the overview page there.
> 
[...]

PR #1666, opened 2025-10-07
Set home page for space

Eshwar1212-maker:
https://github.com/user-attachments/assets/b5df1157-8ed1-47f0-860f-7a30d8107ac8

This PR gives the user and ability to mark any specific page in a space as the home page as requested in this ticket #1646

-A user can mark a page in the space as the homepage with a whole icon that I added in the 'page-header' component.

-Every time a user has a home page for a space it directs them to that page instead of the overview page.

-If a home page is not set the user gets directed to the overview page instead. 

https://github.com/user-attachments/assets/943e59f8-82f3-4773-983c-17d323f5673a.

## Summary by CodeRabbit

* **New Features**
  * You can now set or unset a space’s home page directly from the page header (Home button with tooltip).
* **Improvements**
  * Space cards open directly to the designated home page when one is set, streamlining navigation.
  * More robust handling when updating a space’s home page to prevent invalid selections.

### DIFF
--- a/apps/client/src/features/page/components/header/page-header-menu.tsx
+++ b/apps/client/src/features/page/components/header/page-header-menu.tsx
@@ -1,15 +1,3 @@
-
-
-
-
-
-
-
-
-
-
-
-
 import { ActionIcon, Group, Menu, Text, Tooltip } from "@mantine/core";
 import {
   IconArrowRight,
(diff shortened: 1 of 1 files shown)


Vito0912:
Disclaimer: I am not a maintainer

I love that feature!

But I don't think the button is placed that well. The bar already is overloaded and setting a home page probably isn't going to be an often used action. So imho I think it would better fit in the dot menu instead.

Eshwar1212-maker:
@Vito0912 Yeah thats totally doable to, i thought where it is now is good but we will see what the maintainer says.

PR #1669, opened 2025-10-09
fix: preserve code block language attribute during server-side Markdown import

### LINKED ISSUE #1668: Code block language attribute lost during server-side Markdown import due   to classList incompatibility with happy-dom
## Summary
  When importing Markdown files through the server-side import API, code
  block language identifiers are not preserved. All code blocks end up with
  `language: null` in the database, preventing syntax highlighting and
  Mermaid diagram rendering.

  ## Environment
  - **Docmost version:** 0.23.2
  - **Import method:** Server-side Markdown file import via UI
  - **Database:** PostgreSQL
  - **Node.js version:** 22
  - **Affected parser:** happy-dom (server-side DOM parser)

  ## Reproduction Steps

  1. Create a Markdown file with language-specific code blocks:

  ```markdown
  \`\`\`mermaid
  erDiagram
      CUSTOMER ||--o{ ORDER : places
  \`\`\`

  \`\`\`sql
  SELECT * FROM users WHERE active = true;
  \`\`\`

  \`\`\`javascript
  const greeting = 'Hello World';
  console.log(greeting);
  \`\`\`

  2. Import the file through Docmost UI (Server → Import → Upload Markdown)
  3. Query the database to inspect stored content:

  SELECT content FROM pages WHERE title = 'your-imported-page';

  Expected Behavior

  Code blocks should retain their language attribute in the stored JSON:

  {
    "type": "codeBlock",
    "attrs": {
      "language": "mermaid"
    },
    "content": [
      {"type": "text", "text": "erDiagram\n    CUSTOMER ||--o{ ORDER : 
  places"}
    ]
  }

  Actual Behavior

  All code blocks have language: null regardless of the original language
  specifier:

  {
    "type": "codeBlock",
    "attrs": {
      "language": null
    },
    "content": [
      {"type": "text", "text": "erDiagram\n    CUSTOMER ||--o{ ORDER : 
  places"}
    ]
  }

  Root Cause Analysis

  Import Pipeline

  The Markdown import process follows this chain:

  Markdown → marked.js → HTML → happy-dom → TipTap parseHTML → Database

  The Bug

  The issue originates in @tiptap/extension-code-block's language extraction
   logic:

  // @tiptap/extension-code-block (parent class)
  parseHTML: (element) => {
    const classNames = [...(element.firstElementChild?.classList || [])];
    const languages = classNames
      .filter(className => className.startsWith('language-'))
      .map(className => className.replace(/^language-/, ''));
    return languages[0] || null;
  }

  Problem: The spread operator on classList fails in happy-dom (server-side
  DOM parser). While classList exists as an object, it may not implement the
   iterator protocol correctly, causing the spread operation to return an
  empty array.

  Environment Comparison

  | Environment         | DOM Implementation | [...classList] | Result
         |
  |---------------------|--------------------|----------------|-------------
  -------|
  | Browser             | Native DOM         | ✅ Works        | Language
  extracted |
  | Node.js (happy-dom) | Simulated DOM      | ❌ Returns []   | language: 
  null     |

  This explains why:
  - ✅ Direct editor input: Works (browser environment)
  - ❌ Server-side import: Fails (happy-dom environment)

  Proposed Solution

  Override the parseHTML method in CustomCodeBlock to use getAttribute()
  instead of classList, ensuring compatibility with both browser and
  server-side DOM parsers:

  // packages/editor-ext/src/lib/custom-code-block.ts

  addAttributes() {
    const parentAttrs = this.parent?.() || {};
    const parentLanguage = (parentAttrs as any).language || {};

    return {
      ...parentAttrs,
      language: {
        ...parentLanguage,
        parseHTML: (element) => {
          // Use getAttribute for happy-dom compatibility
          const codeElement = element.firstElementChild;
          if (!codeElement) {
            return null;
          }

          const className = codeElement.getAttribute("class") || "";
          const match = className.match(/language-(\S+)/);
          return match ? match[1] : null;
        },
      },
    };
  }

  Why This Works

  | Approach              | Dependency        | Browser | happy-dom |
  Reliability |
  |-----------------------|-------------------|---------|-----------|-------
  ------|
  | [...classList]        | Iterator protocol | ✅       | ❌         |
  Fragile     |
  | getAttribute("class") | String API        | ✅       | ✅         |
  Robust ✅    |

  The getAttribute("class") method:
  - Returns a standard string ("language-mermaid language-diagram")
  - Works consistently across all DOM implementations
  - Uses regex pattern matching for reliable extraction

  Impact

  Current Impact:
  - ❌ Mermaid diagrams fail to render (display as plain text)
  - ❌ No syntax highlighting for any imported code blocks
  - ❌ Poor user experience when migrating documentation

  After Fix:
  - ✅ All code block languages preserved during import
  - ✅ Mermaid diagrams render correctly
  - ✅ Syntax highlighting works as expected

  Related Components

  - packages/editor-ext/src/lib/custom-code-block.ts - Requires fix
  - apps/server/src/integrations/import/services/import.service.ts - Uses
  marked.js + happy-dom
  - @tiptap/extension-code-block - Parent class with original implementation
  - happy-dom - Server-side DOM parser with limited API support

  Additional Context

  This is a cross-environment compatibility issue. The TipTap codebase was
  designed primarily for browser environments where the DOM API is fully
  implemented. The Docmost import feature uses a server-side DOM parser
  (happy-dom) for performance and security, but this parser has limited
  support for advanced DOM features like iterator protocols on classList.

  The proposed fix maintains backward compatibility while adding robust
  server-side support.

  Testing

  After applying the fix, verify with:

  1. Import a Markdown file with various code block languages
  2. Query the database to confirm language attributes are preserved:
  SELECT jsonb_pretty(content) FROM pages WHERE title = 'test-import';
  3. Verify Mermaid rendering in the UI
  4. Confirm syntax highlighting works for all supported languages

  ---
  Technical Details:
  - Parser: marked v13.0.3
  - DOM: happy-dom (Node.js environment)
  - Editor: TipTap 2.14.0
  - Extension: @tiptap/extension-code-block-lowlight 2.14.0

wangxiao20241123-del:
Fixes #1668 
## Problem

  Code blocks imported from Markdown files lose their language attributes, resulting in `language: null` in the database. This breaks:
  - Mermaid diagram rendering
  - Syntax highlighting for all code blocks

  ## Root Cause

  The parent `CodeBlockLowlight` extension uses `[...(element.firstElementChild?.classList || [])]` to extract language from HTML class names. This approach fails in
  happy-dom (server-side DOM parser) because the spread operator on `classList` doesn't work correctly in Node.js environment.

  **Import pipeline**: Markdown → marked.js → HTML → happy-dom → TipTap parseHTML → Database

  ## Solution

  Override `parseHTML` in `CustomCodeBlock` to use `getAttribute('class')` with regex matching instead of `classList`:

  ```typescript
  const className = codeElement.getAttribute("class") || "";
  const match = className.match(/language-(\S+)/);
  return match ? match[1] : null;
  ```
  Testing

Before Fix
All code blocks have language: null in database
image
After Fix
Language attributes correctly preserved (mermaid, sql, javascript, etc.)
image

## Summary by CodeRabbit

* **New Features**
  * Code blocks now expose a language attribute to improve detection and rendering of programming languages.
* **Bug Fixes**
  * Resolved inconsistent language recognition in code blocks that could cause missing or incorrect syntax highlighting, especially in server-side/Node environments.
  * Language parsing is now more robustly derived from the code element’s class, improving compatibility across browsers and SSR.

### DIFF
--- a/packages/editor-ext/src/lib/custom-code-block.ts
+++ b/packages/editor-ext/src/lib/custom-code-block.ts
@@ -20,6 +20,30 @@ export const CustomCodeBlock = CodeBlockLowlight.extend<CustomCodeBlockOptions>(
       };
     },
 
+    addAttributes() {
+      const parentAttrs = this.parent?.() || {};
+      const parentLanguage = (parentAttrs as any).language || {};
+
+      return {
+        ...parentAttrs,
+        language: {
+          ...parentLanguage,
+          parseHTML: (element) => {
+            // Fix for happy-dom compatibility: extract language from class attribute
+            // instead of using classList which may not work correctly in Node.js environment
+            const codeElement = element.firstElementChild;
+            if (!codeElement) {
+              return null;
+            }
+
+            const className = codeElement.getAttribute("class") || "";
+            const match = className.match(/language-(\S+)/);
+            return match ? match[1] : null;
+          },
+        },
+      };
+    },
+
     addKeyboardShortcuts() {
       return {
         ...this.parent?.(),


Philipinho:
This is no longer an issue in the main branch. We recently fully switched to happy-dom which fixed the issue.
It will be in the next version release.

Let me know if I am wrong.

wangxiao20241123-del:
> This is no longer an issue in the main branch. We recently fully switched to happy-dom which fixed the issue. It will be in the next version release.
> 
> Let me know if I am wrong.

  You're absolutely right! I tested the main branch and confirmed that upgrading TipTap to 2.14.0 has fixed the issue. All code block language attributes are now
  preserved correctly during Markdown import.

  Thank you for the quick response and for already addressing this issue. I'm closing this PR as it's no longer needed.

  For others encountering this issue: please upgrade to the next release which includes TipTap 2.14.0.

PR #1694, opened 2025-10-20
fix: prevent unlicensed user from accessing /security

JosephGKDean:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated access control for the Security page to require both admin privileges and a valid license key. Previously, the page was accessible to admins without an active license.


JosephGKDean:
@coderabbitai generate docstrings

PR #1795, opened 2025-12-16
Fix Notion toggle block import and add support for text/highlight colors

NoeGRANGE:
# Description
This PR addresses issues with the Notion import process and introduces color preservation.

Previously, **Toggle blocks** inside lists were not being imported correctly (only the summary/header was retrieved, while the inner content was lost). I have updated the formatter to correctly traverse details tags and preserve their nested content.

I also fixed **Callout blocks**, which were previously truncated (only the first line was imported). The fix ensures the full content is retrieved, and the associated **emoji/icon** is now correctly imported.

Additionally, I implemented a logic to parse and apply **Text and Highlight colors** from the Notion export. Since Notion uses arbitrary color values, I added a utility to map them to the closest match in our standard color palette.

# Key Changes
1. **Toggle Block Fix:**
- Updated `notionFormatter` to correctly handle `details` elements nested within lists (`ul`/`ol`).
- Ensured that both the summary and the inner content (`div`) of the toggle block are processed and wrapped correctly.

2. **Callout Improvements:**
- **Full Content:** Fixed an issue where multi-line callouts were truncated. Now all lines within the block are preserved.
- **Emoji Support:** Added logic to extract and render the specific emoji associated with the callout block.

3. **Color Support:**
- **Style Parsing:** Added logic to parse the `<style>` block from the HTML export to identify class names associated with colors.
- **Color Mapping:** Implemented a `findClosestColor` algorithm that converts input colors (Hex, RGB, HSL) to HSL space and calculates the distance to find the nearest match in the `TEXT_COLORS` or `HIGHLIGHT_COLORS` palette.
- **Application:** Text colors are applied as `<span>` styles, and background highlights are applied as `<mark>` styles.

# Technical Details
Added helper functions for color conversion (`hslToRgb`, `rgbToHsl`, `parseColor`).

The color matching logic includes a blend-with-white step to handle transparency correctly and avoid mapping transparent colors to black.

Refactored the element wrapping logic to apply color styles down to text nodes and inline elements (like `strong`, `em`, `code`) ensuring consistent styling even inside nested structures.

## Summary by CodeRabbit

* **New Features**
  * Added color palettes and improved color-matching (including grayscale handling) for imported Notion text and highlights.
  * Support for inline math, bookmarks, callouts, to‑do items, lists, and expandable sections with better icon handling and nested color preservation.

* **Improvements**
  * Color formatting is applied earlier and more broadly to ensure accurate propagation and rendering order.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/server/src/integrations/import/utils/import-formatter.ts
+++ b/apps/server/src/integrations/import/utils/import-formatter.ts
@@ -285,6 +285,13 @@ function rgbToHsl(r: number, g: number, b: number): Hsl {
   return { h: h * 360, s, l };
 }
 
+// Weights
+// this is done especially for notions color palette which is more hue focused
+// change const names to better understand their purpose
+const WEIGHT_HUE = 4;
+const WEIGHT_SATURATION = 0.5; // Lower weight for saturation
+const WEIGHT_LIGHTNESS = 0.5; // Lower weight for lightness
+
 function getDistanceHsl(c1: Hsl, c2: Hsl): number {
   let hDiff = Math.abs(c1.h - c2.h);
   if (hDiff > 180) hDiff = 360 - hDiff;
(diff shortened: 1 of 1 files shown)


Philipinho:
Will have a look at this.

PR #1821, opened 2026-01-07
fix: centralize redis config and add TLS support

### LINKED ISSUE #1681: Support for Redis TLS via rediss:// broken due to parseRedisUrl ignoring protocol.
When configuring **Docmost** with a `rediss://` URL (to connect to managed Redis providers such as **Upstash** or **AWS ElastiCache** with TLS enabled), the application fails to start and throws:

```
[ioredis] Unhandled error event: Error: read ECONNRESET
MaxRetriesPerRequestError: Reached the max retries per request limit (20)
```

It seems that the helper function `parseRedisUrl()` used in `RedisConfigService` ignores the `rediss://` protocol and reconstructs the config object without `{ tls: {} }`.

As a result, even if the `REDIS_URL` uses `rediss://`, the Redis client (`ioredis`) attempts a **non-TLS connection** on port `6379`, which the server immediately resets.

---

## ⚙️ Steps to Reproduce

1. Set environment variable:
   ```bash
   REDIS_URL=rediss://default:<token>@real-host-11111.upstash.io:6379
   ```
2. Run the official `docmost/docmost:latest` container.
3. Observe the startup logs (crash after Redis init).
4. In the Upstash dashboard, notice that commands are counted — meaning the TCP connection is established — but the TLS handshake fails.

---

## ✅ Expected Behavior

If `REDIS_URL` starts with `rediss://`, the Redis client should be initialized with:

```ts
new Redis(redisUrl, { tls: {} });
```

so that TLS connections are properly negotiated.

---

## ❌ Actual Behavior

`parseRedisUrl()` returns a plain `host` / `port` / `password` config without TLS, ignoring the protocol.

---

## 🧰 Proposed Fix

Add TLS detection in `RedisConfigService` when building `RedisModuleOptions`:

```ts
const redisConfig = parseRedisUrl(this.environmentService.getRedisUrl());
return {
  readyLog: true,
  config: {
    host: redisConfig.host,
    port: redisConfig.port,
    password: redisConfig.password,
    db: redisConfig.db,
    family: redisConfig.family,
    retryStrategy: createRetryStrategy(),
    tls: redisConfig.protocol === 'rediss:' ? {} : undefined,
  },
};
```

This preserves compatibility for non-TLS Redis setups and enables seamless TLS for `rediss://` connections.

---

## 🗒️ Additional Notes

- The health indicator already works because it uses `new Redis(url)` directly (which respects `rediss://`).
- Only the `RedisConfigService` initialization is affected.

Wessergame:
## What was done
- Introduced `RedisConfigService` providing unified `ioredis` options
- Added TLS support via `REDIS_ENABLE_TLS` or `rediss://` URL
- Reused redis config across:
  - Bull queue
  - WebSocket Redis adapter
  - Collaboration (Hocuspocus)
  - Redis health check
- Removed legacy `parseRedisUrl` and retry helpers

## Why
Fixes incorrect / duplicated Redis configuration and allows proper TLS usage.

## How to test
- Configure `REDIS_URL=rediss://...`
- Optionally set:
  - `REDIS_ENABLE_TLS=true`
  - `REDIS_TLS_REJECT_UNAUTHORIZED`
  - `REDIS_TLS_SERVERNAME`
- Verify:
  - app startup
  - queues
  - websocket collaboration
  - `/health` redis check

## Related issue
Closes https://github.com/docmost/docmost/issues/1681

### DIFF
--- a/apps/server/src/app.module.ts
+++ b/apps/server/src/app.module.ts
@@ -17,6 +17,7 @@ import { ImportModule } from './integrations/import/import.module';
 import { SecurityModule } from './integrations/security/security.module';
 import { TelemetryModule } from './integrations/telemetry/telemetry.module';
 import { RedisModule } from '@nestjs-labs/nestjs-ioredis';
+import { RedisConfigModule } from './integrations/redis/redis-config-module';
 import { RedisConfigService } from './integrations/redis/redis-config.service';
 
 const enterpriseModules = [];
@@ -39,7 +40,12 @@ try {
     DatabaseModule,
     EnvironmentModule,
     RedisModule.forRootAsync({
-      useClass: RedisConfigService,
+      imports: [RedisConfigModule],
+      inject: [RedisConfigService],
+      useFactory: (redisConfig: RedisConfigService) => ({
+        readyLog: true,
+        config: redisConfig.getOptions(),
+      }),
     }),
     CollaborationModule,
     WsModule,
--- a/apps/server/src/collaboration/collaboration.gateway.ts
+++ b/apps/server/src/collaboration/collaboration.gateway.ts
@@ -6,25 +6,21 @@ import { PersistenceExtension } from './extensions/persistence.extension';
 import { Injectable } from '@nestjs/common';
 import { Redis } from '@hocuspocus/extension-redis';
 import { EnvironmentService } from '../integrations/environment/environment.service';
-import {
-  createRetryStrategy,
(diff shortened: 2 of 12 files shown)


iavael:
@Philipinho can you please look into this PR?

Philipinho:
Thanks for you work.
TLS was fixed here https://github.com/docmost/docmost/pull/2021.

PR #1826, opened 2026-01-09
feat: TipTap v3 migration finalization

areknawo:
- Move from Tippy.js to Floating UI;
- Manually set `isInitialized` to force immediate rendering of React node views;
- Update collaboration connection logic for HocusPocus v3


Philipinho:
Good work.
It seems to work nicely, and the unnecessary react NodeView flicker is no more with the `isInitialized` hack. 

I noticed the deprecated provider function error in the console log. You may want to have a look at its usage.
```
HocuspocusProvider::connect() is deprecated and does not do anything. Please connect/disconnect on the websocketProvider, or attach/deattach providers.```

PR #1844, opened 2026-01-15
fix: collab status display

Philipinho:


Philipinho:
Merged in tiptap3-migration branch.

PR #1845, opened 2026-01-15
feat: add per-page control for public PDF export

### LINKED ISSUE #1829: Feat : Allow Export Pdf in Public Shared link
While Sharing a page with a public link ,

Add a feature:
Allow to export pdf - toggle (on/off) - Default (off)

It will help the shared one to make copy .

Image

HeeraRana247453:
## What does this PR do?

Adds a per-page toggle to control whether PDF export is allowed on public shared pages.

## Why is this needed?

Currently, any publicly shared page can be exported as a PDF.
For sensitive or read-only documentation, this can be undesirable even when view access is intentionally public.

This change gives page owners explicit control over whether PDF export should be allowed.

## What’s included?

- New `allowPublicPdfExport` flag on shares (default: **false**)
- Public PDF export is blocked when the toggle is disabled
- “Print PDF” option is hidden on public pages when disabled
- No changes to private pages or authenticated users

## Implementation notes

- Existing PDF export logic is reused
- Default behavior remains secure and backward-compatible
- No impact on current permissions or roles

## Related issue

Closes #1829


HeeraRana247453:
Hi 👋  
Just a gentle follow-up in case this slipped through.

All checks are passing and the change is intentionally small and backward-compatible.  
Happy to make any adjustments if needed.

Thanks for your time!

Philipinho:
Thanks for your contribution.
I unfortunately wont be merging this PR.

1. It does not fix the issue.
2. You deleted server code you shouldn't have.
3. This is not a feature we are looking to introduce yet.

HeeraRana247453:
Thanks for the feedback.  
Understood the concerns and project direction.

Closing this from my side.  
Appreciate the time and review.

PR #1851, opened 2026-01-19
feat: Improve placeholder and upload handling for assets

areknawo:
- Refactor asset (image, video, attachment) upload functions to insert nodes immediately;
- Add suppot for placeholder modes to asset views that matching the end block dimensions to reduce flickering;

### DIFF
--- a/apps/client/public/locales/en-US/translation.json
+++ b/apps/client/public/locales/en-US/translation.json
@@ -328,6 +328,8 @@
   "Upload any image from your device.": "Upload any image from your device.",
   "Upload any video from your device.": "Upload any video from your device.",
   "Upload any file from your device.": "Upload any file from your device.",
+  "Uploading {{name}}": "Uploading {{name}}",
+  "Uploading file": "Uploading file",
   "Table": "Table",
   "Insert a table.": "Insert a table.",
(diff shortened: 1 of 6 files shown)


packages/editor-ext/src/lib/image/image-upload.ts
```diff
@@ -1,127 +1,111 @@
...
-      tr.setMeta(uploadKey, {
-        add: {
-          id,
-          pos,
-          src: reader.result,
-        },
-      });
 
-      insertTrailingNode(tr, pos, view);
+    const imageDimensions = imageDimensionsFromData(await file.bytes());
```
Philipinho: This returns
```
image-upload.ts:36 Uncaught (in promise) TypeError: file.bytes is not a function
    at image-upload.ts:36:64
    at input.onchange (menu-items.ts:178:15)
```
Changing `bytes()` to `stream()` fixes it.
```
const imageDimensions = await imageDimensionsFromStream(file.stream());
```

areknawo: Thanks! Looks like `bytes()` isn't widely-supported just yet: https://developer.mozilla.org/en-US/docs/Web/API/Blob/bytes

PR #1887, opened 2026-01-31
fix(client): ensure sidebar remains visible on shared subpages

MatheusLuisLorscheiter:


Philipinho:
Thanks for contributing.

PR #1912, opened 2026-02-06
feat(ee): AI menu

areknawo:
This PR adds an AI menu with a list of predefined commands and Insert/Discard-based UX.

localhost_5173_s_general_p_hello-worl-TqCV7iqjin(Macbook Pro 14_)

### DIFF
--- a/apps/client/src/ee/ai/services/ai-service.ts
+++ b/apps/client/src/ee/ai/services/ai-service.ts
@@ -43,13 +43,16 @@ export async function generateAiContentStream(
     }
 
     const processStream = async () => {
+      let buffer = "";
       try {
         while (true) {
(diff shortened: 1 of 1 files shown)


Philipinho:
*Bug*

- Open the AI menu without closing it
- Navigate to a different pages, the menu remains open.

Philipinho:
Could we make the floating menu smarter so it don't overlap like this?
image

areknawo:
> Could we make the floating menu smarter so it don't overlap like this? image

Done! I've turned-off the auto-filpping, and also switched the menu to absolute positioning. This way, when the menu is used closer to the bottom, the scrollable area just expands for additional space.

PR #1922, opened 2026-02-09
fix: resolve keystroke input being swallowed after link in Firefox

### LINKED ISSUE #1748: Cannot type before link when link is the first element of a line
Hello, 
When a link is the first element of a line, it seems that I cannot type text before the link. 

https://github.com/user-attachments/assets/88664cc6-4969-4d9b-88c1-57cf2cade6b9

Image

I am running `docmost/docmost:0.24.0-alpha-1` with license. 

(Side note, I would really appreciate a "plain markdown editor" like this one in Github)

### LINKED ISSUE #1773: [BUG] All keystrokes ignored when adding text after a link in some conditions
How to reproduce "All keystrokes ignored":
1. start a new line
2. type `foo bar,`
3. then go back before the comma, enter space then `(`
4. paste a link
5. try to close the parenthesis, it won't do anything no matter how many times you punch the ) key.

olivierlambert:
In Firefox, when the cursor is at the right boundary of a link mark, contenteditable inserts new text inside the <a> element. ProseMirror then rejects the DOM mutation because the link mark has inclusive: false, causing keystrokes to be silently swallowed. Unlike Chrome, Firefox also does not fire ProseMirror's handleTextInput callback in this state.

This adds a ProseMirror plugin that intercepts printable character keydowns at link mark boundaries and programmatically inserts the text without the link mark, bypassing Firefox's native contenteditable behavior entirely.

Fixes #1773

Note: functionally tested. Before: bug visible only in Firefox. Now the bug is fixed in Firefox without any regression on Chrome.

edit: now improved via a 2nd commit to fix #1748 at the same time.

### DIFF
--- a/apps/client/src/features/editor/components/bubble-menu/link-selector.tsx
+++ b/apps/client/src/features/editor/components/bubble-menu/link-selector.tsx
@@ -2,6 +2,7 @@ import { Dispatch, FC, SetStateAction, useCallback } from "react";
 import { IconLink } from "@tabler/icons-react";
 import { ActionIcon, Popover, Tooltip } from "@mantine/core";
 import { useEditor } from "@tiptap/react";
+import { TextSelection } from "@tiptap/pm/state";
 import { LinkEditorPanel } from "@/features/editor/components/link/link-editor-panel.tsx";
 import { useTranslation } from "react-i18next";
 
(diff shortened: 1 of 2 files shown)


Philipinho:
Thanks for contributing.

PR #1924, opened 2026-02-09
fix: cursor jumps to end of text when editing a comment

### LINKED ISSUE #1791: [BUG] Issue while editing comment, cursor goes to end of text at each keystroke
I am using Firefox on Linux x86_64, on sway/wayland.

how to reproduce : 
1. create a page
2. write and submit a comment
3. edit the comment
4. put the cursor in the middle of the text and try to add text: at each keystroke it will print the first character a the correct place then move the cursor to the end of the text.

This makes it almost impossible to edit your comment in the middle, you can only append text at the end.
Or else you need to delete the text and re-type it.

olivierlambert:
When editing a comment mid-text, the cursor would jump to the end after every keystroke, making it impossible to insert text at any position other than the end.

Root cause: on each keystroke, the comment editor's onUpdate callback updated parent state (setContent), which changed the defaultContent prop passed back to CommentEditor. A useEffect watching defaultContent then called commentEditor.commands.setContent(), which reset the entire editor content and moved the cursor to the end.

Fix:
- Store in-progress edits in a ref instead of state to avoid triggering React re-renders and the prop->effect->setContent cascade
- Read from the ref when saving the comment
- Sync the ref back into state after a successful save so the read-only view updates immediately
- Guard the setContent useEffect to only run for read-only editors, so websocket-driven updates from other browsers still work

Fixes #1791

Functionally tested on Firefox and Chrome: mid-text editing, saving, cross-browser live updates via websocket.

### DIFF
--- a/apps/client/src/features/comment/components/comment-list-item.tsx
+++ b/apps/client/src/features/comment/components/comment-list-item.tsx
@@ -133,6 +133,7 @@ function CommentListItem({
     setIsEditing(true);
   }
   function cancelEdit() {
+    editContentRef.current = null;
     setIsEditing(false);
   }
 


Philipinho:
Thanks for contributing.

PR #1953, opened 2026-02-18
fix: redirect to original page after re-authentication

julien-f:
## Summary

- When a session expires, the current URL is now preserved as a `redirect` query parameter on the login page
- After successful login (including MFA and LDAP flows), the user is redirected back to their original page instead of always landing on `/home`
- Includes open redirect protection (rejects protocol-relative URLs)

## Test plan

- [ ] Navigate to a specific page, clear auth cookie, trigger a request → verify redirect to `/login?redirect=<original-path>`
- [ ] Log back in → verify you land on the original page
- [ ] Verify MFA flow preserves the redirect param through `/login/mfa`
- [ ] Verify navigating directly to `/login` (no redirect param) still goes to `/home` after login
- [ ] Verify `?redirect=//evil.com` is rejected and falls back to `/home`

### DIFF
--- a/apps/client/src/ee/components/ldap-login-modal.tsx
+++ b/apps/client/src/ee/components/ldap-login-modal.tsx
@@ -7,7 +7,7 @@ import { notifications } from "@mantine/notifications";
 import { useNavigate } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import { IAuthProvider } from "@/ee/security/types/security.types";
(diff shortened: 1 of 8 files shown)


Philipinho:
I unfortunately couldn't push my changes to this branch since it is your `main`.
I opened a new PR with a security fix https://github.com/docmost/docmost/pull/1959.
The current check doesn't prevent redirects to urls like `?redirect=/\evil.com`.

Thanks for contributing.

PR #1954, opened 2026-02-18
fix: allow deleting last character in headings

julien-f:
## Summary

- The copy-link decoration widget (`contentEditable="false"`) injected inside headings prevented browsers from deleting the last remaining character via Backspace or Delete keys
- Fix: only render the widget when the heading has more than one character of content (`node.content.size > 1` instead of `> 0`)

## Test plan

- [ ] Create a heading (e.g. type `# A`)
- [ ] Place cursor after the single character and press Backspace — should delete it
- [ ] Create a heading, place cursor before the single character and press Delete — should delete it
- [ ] Verify the copy-link button still appears on headings with 2+ characters


Philipinho:
Thanks for contributing.

PR #2022, opened 2026-03-15
fix: pin jose@4 for openid-client to prevent OIDC breakage

michidk:
## Summary

`openid-client@5` requires `jose@^4.15.5 || ^5.0.0`, but `@modelcontextprotocol/sdk` (added in #1976) depends on `jose@6`. With `shamefully-hoist = true` in `.npmrc`, pnpm hoists `jose@6` to the root `node_modules/`, which `openid-client` then resolves at runtime.

`jose@6` has breaking API changes that cause OIDC token exchange to fail silently.

## Fix

Scoped pnpm override in `package.json`:

```json
"openid-client>jose": "4.15.9"
```

This ensures `openid-client` always resolves `jose@4` regardless of hoisting, without affecting other packages that legitimately use `jose@6`.

## Why not upgrade openid-client to v6?

The proper long-term fix is upgrading `openid-client` from v5 to v6 (which uses `jose@6` natively), but v5→v6 is a full API rewrite (class-based → functional). Since the OIDC usage lives in the EE submodule, this override is the safe interim fix. It can be removed once `openid-client` is upgraded to v6+.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
       "@tiptap/core": "patches/@tiptap__core.patch"
     },
     "overrides": {
+      "openid-client>jose": "4.15.9",
       "jsdom": "25.0.1",
       "jsonwebtoken": "9.0.3",
       "prosemirror-changeset": "2.3.1",


Philipinho:
Could you share what issues you ran into with more detail?

michidk:
We run a fork with a custom OIDC implementation using `openid-client@5` (`Issuer.discover()`, `client.callback()`, `tokenSet.claims()`). After `@modelcontextprotocol/sdk` was added in #1976, OIDC logins started failing at the callback stage.

**Root cause:** `@modelcontextprotocol/sdk` depends on `jose@6`. With `shamefully-hoist = true` in `.npmrc`, pnpm hoists `jose@6` to the root `node_modules/`. Since `openid-client@5` specifies `jose@^4.15.5 || ^5.0.0`, it resolves the hoisted `jose@6` at runtime instead of its compatible version. `jose@6` has breaking API changes (class-based → function-based), so `openid-client`'s internal calls to jose fail.

**Symptom:** The OIDC login redirect to the IdP works fine, but the callback (`client.callback()` — token exchange and JWT verification) throws. In our case it was caught and redirected the user back to `/login?error=oidc_failed`. Without explicit error logging it can appear to fail silently.

**Note:** This affects upstream too — the EE OIDC module in `apps/server/src/ee/` also uses `openid-client@5` and would hit the same issue.

Philipinho:
Thanks for the detail.

I can not reproduce on my end, and never had any such report from enterprise users.
This is likely an issue with your fork and implementation.

Anyways, we will upgrade to openid-client v6 soon, as there are fewer CJS and ESM compatibility concerns with recent Node.js versions.

PR #2023, opened 2026-03-15
feat: add Gravatar fallback for user avatars

michidk:
Adds Gravatar as a fallback when users have no uploaded avatar. Instead of always showing colored initials, the avatar component now checks Gravatar first using the user's email hash.

## How it works

1. **New utility** (`apps/client/src/lib/gravatar.ts`): `getGravatarUrl(email)` hashes the email with MD5 (pure-JS, no dependencies) and returns a Gravatar URL with `d=404`.

2. **CustomAvatar** (`apps/client/src/components/ui/custom-avatar.tsx`): Accepts an optional `email` prop. When no `avatarUrl` is set, falls back to the Gravatar URL. The `d=404` parameter means Gravatar returns a 404 when no image exists, which lets Mantine's `<Avatar>` fall back to the colored initials as before.

3. **Fallback chain**: Uploaded avatar → Gravatar → Colored initials

## Changes

- **New file**: `apps/client/src/lib/gravatar.ts` — MD5 hash + Gravatar URL builder
- **Modified**: `apps/client/src/components/ui/custom-avatar.tsx` — added `email` prop + Gravatar fallback
- **Modified** (9 files): Threaded `email` prop to `CustomAvatar` in all user-facing components where email is available in the data model (user-info, top-menu, comments, group members, space members, workspace members, multi-select dropdowns)

## Notes

- Some `CustomAvatar` call sites (notifications, share list, page history, audit logs) don't have email in their API response types, so those are left unchanged — they continue to use initials as fallback.
- The MD5 implementation is a compact pure-JS version (~200 lines) to avoid adding a dependency. An alternative would be to use `crypto.subtle` but it doesn't support MD5 (only SHA family), so a JS implementation is necessary for Gravatar compatibility.

### DIFF
--- a/apps/client/src/components/common/user-info.tsx
+++ b/apps/client/src/components/common/user-info.tsx
@@ -10,7 +10,7 @@ interface UserInfoProps {
 export function UserInfo({ user, size }: UserInfoProps) {
   return (
     <Group gap="sm" wrap="nowrap">
-      <CustomAvatar avatarUrl={user?.avatarUrl} name={user?.name} size={size} />
+      <CustomAvatar avatarUrl={user?.avatarUrl} name={user?.name} email={user?.email} size={size} />
       <div>
         <Text fz="sm" fw={500} lineClamp={1}>
           {user?.name}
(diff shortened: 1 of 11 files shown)


michidk:
So we could also pull in a dependency (like https://github.com/sindresorhus/gravatar-url) instead of calculating the Gravatar URL ourselves. In my fork, I did it the custom way to prevent having to pull in another dep, but I'm open to changing this if this feature is interesting for you.

Philipinho:
No plans to support a third-party avatar system.

PR #2024, opened 2026-03-15
feat: add HTML source editor mode

michidk:
Adds a toggleable source mode to the page editor that lets users view and edit the page's HTML directly. Useful for debugging, fixing formatting issues, or precise content editing that's hard to do in WYSIWYG mode.

I implemented it because it was not possible to change the URL for an already existing image. I could only remove it and paste it again.
Also, my first iteration was a markdown editor, but some elements like tables might have custom styling to it - so editing it in markdown would lead to information loss. Therefore, it's now a raw HTML editor.

image

## How it works

- Click the `</>` (code) icon in the page header to toggle between rich-text and source mode
- Source mode shows the TipTap HTML in a CodeMirror editor with syntax highlighting, line numbers, code folding, and bracket matching
- A lightweight HTML pretty-printer formats the raw TipTap output for readability (indentation around block elements)
- On switching back, the edited HTML is applied directly to the TipTap editor via `setContent()`
- The rich-text editor stays mounted (hidden via CSS) while in source mode so the Tiptap/Y.js instance remains alive
- Source mode resets automatically when navigating to a different page

## Changes

**New files:**
- `apps/client/src/features/editor/source-editor.tsx` — CodeMirror-based HTML source editor with pretty-printer, theme matching, and proper cleanup on unmount
- `apps/client/src/features/editor/styles/source-editor.module.css` — Wrapper styles with responsive padding

**Modified files:**
- `apps/client/package.json` — Added `@uiw/react-codemirror` and `@codemirror/lang-html` dependencies
- `apps/client/src/features/editor/atoms/editor-atoms.ts` — Added `sourceModeAtom`
- `apps/client/src/features/editor/full-editor.tsx` — Conditionally renders SourceEditor; keeps PageEditor mounted but hidden in source mode
- `apps/client/src/features/page/components/header/page-header-menu.tsx` — Added source toggle button (`IconCode`) next to the page state control

## Design decisions

- **HTML, not Markdown**: Editing HTML directly avoids lossy markdown round-trips (not all TipTap node types have clean markdown equivalents). What you see is what TipTap stores.
- **CodeMirror**: Provides syntax highlighting, folding, bracket matching, and search out of the box — much better UX than a plain textarea for editing HTML.
- **PageEditor stays mounted**: Unmounting destroys the Tiptap instance, which means `setContent()` on switch-back would silently fail. Hiding with `display: none` keeps Y.js alive.
- **Uncontrolled CodeMirror**: Uses `initialValue` + `onChange` ref pattern instead of a controlled `value` prop, so Y.js editor recreation doesn't clobber user edits mid-session.
- **Only shown for editable pages**: The toggle button is hidden on read-only pages.

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -55,6 +55,8 @@
     "semver": "^7.7.3",
     "socket.io-client": "^4.8.3",
     "tiptap-extension-global-drag-handle": "^0.1.18",
+    "@codemirror/lang-html": "^6.4.11",
+    "@uiw/react-codemirror": "^4.25.8",
     "zod": "^4.3.6"
   },
   "devDependencies": {
--- a/apps/client/src/features/editor/atoms/editor-atoms.ts
+++ b/apps/client/src/features/editor/atoms/editor-atoms.ts
@@ -9,4 +9,6 @@ export const readOnlyEditorAtom = atom<Editor | null>(null);
 
 export const yjsConnectionStatusAtom = atom<string>("");
 
+export const sourceModeAtom = atom<boolean>(false);
+
 export const showAiMenuAtom = atom(false);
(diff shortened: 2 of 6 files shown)


Philipinho:
No pans to support this at this time.

PR #2039, opened 2026-03-20
fix: notification items are now real links

julien-f:
## Summary

- Notification items in the notification panel were rendered as `<button>` elements with a `useNavigate()` click handler, making it impossible to open them in a new tab or window.
- Replaced `UnstyledButton` (button) with `UnstyledButton component={Link}` so each item renders as a real `<a>` element — the browser handles all navigation natively.
- Regular left-click still uses SPA navigation and closes the popover.
- Ctrl/Cmd/Shift+click and middle-click open the page in a new tab; all of these mark the notification as read.
- Right-click (context menu) does **not** mark as read.
- Added `display: block` to the CSS class to fix layout regression caused by `<a>` defaulting to `display: inline` (unlike `<button>`).

## Notes

> ⚠️ This code was written by an AI assistant (Claude) and has been manually tested by a human. Please review carefully before merging.

## Test plan

- [ ] Left-click a notification — navigates in-place, marks as read, closes popover
- [ ] Ctrl/Cmd+click — opens in new tab, marks as read
- [ ] Middle-click — opens in new tab, marks as read
- [ ] Right-click — opens context menu, does **not** mark as read
- [ ] Notification items display correctly (no extra spacing, hover highlight is a single block)

### DIFF
--- a/apps/client/src/features/notification/components/notification-item.tsx
+++ b/apps/client/src/features/notification/components/notification-item.tsx
@@ -13,7 +13,7 @@ import {
 import { CustomAvatar } from "@/components/ui/custom-avatar";
 import { INotification } from "../types/notification.types";
 import { Trans, useTranslation } from "react-i18next";
-import { useNavigate } from "react-router-dom";
+import { Link } from "react-router-dom";
 import { useState } from "react";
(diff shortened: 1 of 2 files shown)


Philipinho:

PR #2050, opened 2026-03-25
fix: strip trailing newlines from pasted text to prevent list breakage

### LINKED ISSUE #2049: bullet point list behaving very un-naturally after pasting text
how to reproduce (tested in both Firefox and Chromium):

1/ create a bullet point
2/ paste the output of, say, `uname -a` command in it
3/ it will paste the text on the currently focused bullet point line and put your cursor on the next line, without creating it as a new bullet point
4/ then say you want to create a new bullet point. You then delete this empty line to go back to previous bullet line in order to press "enter" so that a new bullet point is created
5/ you then press backspace to do that

what happens is that instead of removing the newline and putting the cursor on the previous line (which is your first bullet), it will stay on the blank newline and actually ***delete*** the bullet point formatting of the previous line.

What I would expect instead:
2 things
- the paste, if it creates a newline, should create a new bullet as well, not a blank line (but I'm pretty sure I didn't select/copy any "newline" in my terminal so I would even not expect the new line to be created in the first place).
- if pressing backspace on the blank line, it should ***delete*** the currently selected blank line and go back to the bullet line right before it and not remove its bullet formatting

I can't record my screen right now, tell me if the bug description is not clear enough.

olivierlambert:
## Summary

- Strip trailing newlines from pasted text before markdown conversion to prevent broken list behavior
- Terminal output often ends with `\n`, which the markdown parser (with `breaks: true`) converts to a `<br>`, creating an empty line outside the list item
- This caused backspace on the empty line to incorrectly remove the previous bullet's formatting instead of joining lines

Fixes #2049

## Test plan

- [x] Create a bullet point in the editor
- [x] Paste terminal output (e.g. `uname -a`) into the bullet
- [x] Verify no extra empty line is created after the pasted text
- [x] Verify pressing Enter after the pasted text creates a new bullet point
- [x] Verify pasting markdown content from VS Code still works correctly

### DIFF
--- a/apps/client/src/features/editor/extensions/markdown-clipboard.ts
+++ b/apps/client/src/features/editor/extensions/markdown-clipboard.ts
@@ -1,7 +1,7 @@
 // adapted from: https://github.com/aguingand/tiptap-markdown/blob/main/src/extensions/tiptap/clipboard.js - MIT
 import { Extension } from "@tiptap/core";
 import { Plugin, PluginKey } from "@tiptap/pm/state";
-import { DOMParser } from "@tiptap/pm/model";
+import { DOMParser, Fragment, Slice } from "@tiptap/pm/model";
 import { find } from "linkifyjs";
 import { markdownToHtml } from "@docmost/editor-ext";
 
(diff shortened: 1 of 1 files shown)


fallen:
I was able to test this PR with my usual reproducer:
* create a bullet point
* paste the output of `uname -a`

Now with this PR, after pasting, it does not create the empty newline anymore and I don't have the bug.
I can then hit enter to continue adding new bullet point items.

:+1:

olivierlambert:
@Philipinho we split the PR code (done on my side) with the functional testing (done @fallen side), and we are happy so say that https://github.com/docmost/docmost/issues/2049 is fixed with this :+1:

Philipinho:
Thanks for contributing.

PR #2052, opened 2026-03-25
fix: ctrl-k behavior on macOS

notpeter:
Fixes a bug where on macOS ctrl-k does two things (1) cuts to the end of the line (2) triggers the search prompt. I changed it so that search is explicitly cmd-k on macOS/iOS and ctrl-k everywhere else, which fixes `ctrl-k` on macOS.

I also:
- Made it so the search bar shows the `cmd-k` suggestion on Apple machines
- Narrowed the no-op `ctrl-s` / `cmd-s` to be platform specific -- I don't think `win+s` (Windows Search) would've made it to the browser anyways.

Tested locally and it seemed to work as expected.
Screenshot 2026-03-25 at 12 53 40 PM

AI Notice: This PR was Claude Code assisted.

### DIFF
--- a/apps/client/src/features/comment/components/comment-editor.tsx
+++ b/apps/client/src/features/comment/components/comment-editor.tsx
@@ -10,6 +10,7 @@ import { useTranslation } from "react-i18next";
 import EmojiCommand from "@/features/editor/extensions/emoji-command";
 import mentionRenderItems from "@/features/editor/components/mention/mention-suggestion";
 import MentionView from "@/features/editor/components/mention/mention-view";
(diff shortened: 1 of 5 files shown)


Yundi339:
Hi, Does this also need to be modified?
 comment-editor.tsx:86 Ctrl+Enter to submit a comment

notpeter:
> Hi, Does this also need to be modified? comment-editor.tsx:86 Ctrl+Enter to submit a comment

Good catch. Changed there too.

Philipinho:
Thanks for contributing.

Issue #2047, opened 2026-03-25
Missing database indexes on critical FK columns cause excessive sequential scans


hwsong:
## Summary

Several core tables in Docmost lack indexes on frequently queried foreign key columns. This causes PostgreSQL to perform full sequential scans on every API request, significantly degrading performance as data grows.

## Environment

- **Docmost version**: 0.70.3
- **PostgreSQL**: 16.10 (CloudNativePG HA, 3 instances)
- **Configuration**: `shared_buffers=1GB`, `max_connections=100`
- **Scale**: ~5,800 pages, ~170 users, ~3,600 attachments, ~30 spaces

## Problem

After analyzing `pg_stat_user_tables`, we found that nearly all critical tables rely entirely on sequential scans due to missing indexes:

| Table | Seq Scan % | Seq Scans | Tuples Read via Seq Scan | Actual Rows |
|-------|-----------|-----------|-------------------------|-------------|
| `page_history` | **99.7%** | 28,601 | 70,310,741 | 3,825 |
| `group_users` | **99.3%** | 1,016,392 | 25,705,658 | 711 |
| `space_members` | **99.8%** | 443,796 | 23,960,415 | 96 |
| `users` | **98.7%** | 495,666 | 8,410,935 | 172 |
| `pages` | 33.1% | 255,069 | 1,674,357,721 | 5,769 |
| `attachments` | 28.3% | 27,245 | 197,928,447 | 3,633 |
| `backlinks` | 47.1% | 26,654 | 44,989,058 | 414 |

The `group_users` and `space_members` tables are particularly impactful because they are queried on **every single API request** during permission/ability checks (e.g., `spaceAbility.createForUser()` → scans `space_members` by `user_id`, then `group_users` by `user_id`).

The `pages` table is especially expensive — 1.67 billion tuples read via sequential scan with only 5,769 rows — because its TOAST storage (holding `ydoc` bytea and `content` jsonb) makes each full scan very costly.

## Root Cause

The migration files create tables with FK columns but **do not create indexes on those FK columns**. PostgreSQL (unlike MySQL) does not automatically create indexes on foreign key columns.

### Migration audit

| Table | Migration | Indexes Created | Missing Indexes |
|-------|-----------|----------------|-----------------|
| `page_history` | `20240324T086400` | PK only | `page_id` |
| `group_users` | `20240324T085700` | PK, unique(group_id, user_id) | `user_id` (standalone) |
[...]

PR #2071, opened 2026-03-29
fix: add new indexes for better performance

Philipinho:
Prior PR https://github.com/docmost/docmost/pull/1832.


Philipinho:
Closes https://github.com/docmost/docmost/issues/2047.

PR #2087, opened 2026-04-03
feat: Basic Summary Row Functions Support in Tables

### LINKED ISSUE #2012: Feature Request: Basic Formula Support in Tables (e.g., SUM)
### Description

It would be very useful if tables in Docmost could support simple formulas, especially basic arithmetic operations such as sums.

Currently, tables work well for displaying structured information, but when using them for small calculations (for example budgets, equipment lists, or cost overviews), the lack of formulas means that totals must be calculated manually.

### Proposed Functionality

Support for simple formulas inside table cells, for example:

- `SUM` of a column or selected cells  
- Basic arithmetic operations (`+`, `-`, `*`, `/`)  
- Optional: referencing other cells within the table

### Example Use Case

A simple cost overview:

| Item   | Quantity | Price | Total |
|--------|----------|-------|-------|
| Camera | 2        | 500   | `=Quantity * Price` |
| Lens   | 1        | 800   | `=Quantity * Price` |
| **Sum** |          |       | `=SUM(Total)` |

### Why This Would Help

This would turn tables into lightweight calculation tools without needing a full spreadsheet system.
Even limited support (for example only `SUM` and basic arithmetic) would already cover many common use cases.

### Additional Note

The request is specifically for *simple* formulas, not a full spreadsheet engine. Keeping the implementation minimal would likely already provide significant value.

Tarantula471:
[feat: Basic Summary Row Functions Support in Tables](https://github.com/docmost/docmost/commit/965ae848a0ec6d857b00484ad898c6a930017af3)

Fixes #2012


Philipinho:
No plans to support this in normal tables.

PR #2118, opened 2026-04-18
feat(tree): Auto-expand current page in sidebar tree

Yundi339:
Hi, I want to add this function of automatically expanding the current page, which can be much more convenient.
I originally wanted to fully expand the function, but I found that the current loading strategy is to load on demand, so in order to maintain consistency, I chose to only expand the current subdirectory, and the next level will not be automatically expanded.


Yundi339:
I also have an idea to display how many subpages there are next to each page, but I found that this may cause permission issues and make changes more difficult. Is there any good solution?

Philipinho:
Closing as the tree system was rebuilt.
We added a new `*` shortcut to expand root pages.

PR #2123, opened 2026-04-20
fix: prevent browser tab fallback in editor

### LINKED ISSUE #2005: [BUG] Using tab in a bullet list jumps to search box in code block
I have a page with a code block and many bullet points. For some of these bullet points, when I try to press tab to make this a sub bullet, instead of making this change the page jumps to the search box in the code block which is used to select the language. This happens every time and I have not yet found a way around this.

mayhemking007:
### Problem

Pressing `Tab` inside the editor sometimes triggers the browser's default behavior (e.g., focusing the code block language selector) instead of staying within the editor.

This happens when list indentation is not applicable (e.g., first bullet or invalid nesting).

### Root Cause

If no extension handles the `Tab` key, ProseMirror falls back to the browser’s default behavior.

### Behavior

List items can only be indented when a valid preceding sibling exists. In cases where indentation is not possible (e.g., first bullet or invalid hierarchy), the expected behavior is to do nothing.

Previously, these no-op cases triggered browser tab navigation. This fix ensures the editor retains focus and performs no action instead.

### Solution

Prevent the browser default behavior and delegate handling back to ProseMirror:

```ts id="trimmed1"
event.preventDefault();

return editor.view.someProp("handleKeyDown", (f) =>
  f(editor.view, event)
);
```

This ensures:

* Existing behaviors (list indentation, table navigation, etc.) continue to work
* No-op cases do nothing instead of triggering browser focus changes

### Testing

Tested the following cases:

* Bullet list indentation (valid case)
* First bullet (no indent case)
* Nested lists
* Lists near code blocks
* Table navigation using Tab
* Code block tab behavior
* Shift + Tab behavior

All behaviors remain correct and no browser fallback occurs.

### Demo
Quick demo showing Tab behavior before and after fix:

#### Before -

https://github.com/user-attachments/assets/d089da85-f73c-4657-badc-323a39283e12

#### After-

https://github.com/user-attachments/assets/8e7f0ccb-de45-4ca3-9d69-329dfdfb45ff

Fixes #2005


Philipinho:
Thanks for contributing.

Philipinho:
Reverted as this is required to meet accessibility standards.

mayhemking007:
Thanks for the clarification, makes sense regarding accessibility.

Just to clarify the expected behavior here: when sinkListItem is not applicable (e.g. first bullet / invalid nesting), should Tab follow normal browser focus navigation, or should it remain in the editor and simply perform no action?

Philipinho:
@mayhemking007 The user should be able to exit the editor or focus on the next focusable element using the keyboard Tab key.

PR #2131, opened 2026-04-27
fix(i18n): Correct German column count label rendering

### LINKED ISSUE #2130: Missing label for column layout dropdown
When using the column layout feature, the dropdown for selecting column configurations (e.g. {count, plural, one {# Spalte} other {# Spalten}}) has no visible label or description.

docmost 0.80.0
Language: German

Image

mayhemking007:
## Fix German column count label rendering

The German translation for column count was using ICU plural syntax:
```
{count, plural, one {# Spalte} other {# Spalten}}
```
However, the current i18n setup (react-i18next) does not support ICU formatting, causing the raw string to be displayed in the UI.

## Fix
Replaced ICU syntax with standard i18next interpolation:
```
"{{count}} Spalten"
```

Since the UI only supports 2–5 columns, plural form is always correct.

## Result
Column count now renders correctly in German:
- 2 Spalten
- 3 Spalten
- etc.

### Before Fix - 

Screenshot 2026-04-27 190616

### After Fix - 

Screenshot 2026-04-27 192256

Fixes #2130

### DIFF
--- a/apps/client/public/locales/de-DE/translation.json
+++ b/apps/client/public/locales/de-DE/translation.json
@@ -370,7 +370,7 @@
   "Write anything. Enter \"/\" for commands": "Schreiben Sie irgendetwas. Geben Sie \"/\" für Befehle ein",
   "Write...": "\"Schreiben...\"",
   "Column count": "Spaltenanzahl",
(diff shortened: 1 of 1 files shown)


Philipinho:
Thanks for contributing.

PR #2161, opened 2026-05-07
Add alias for /toc and /ol

notpeter:
Make `/toc` resolve to "Subpages (Child Pages)"
Make `/ol` resolve to "Numbered List"


Philipinho:
Thanks for contributing.

PR #2165, opened 2026-05-08
fix: sync html lang with user locale for spellcheck

### LINKED ISSUE #1663: Bug: HTML language static in English affects spell checking
In the `apps/client/index.html` file, the HTML language is statically set to English, even if you've specified a different language in the settings.

It would be great if the language in the HTML could be changed to the configured language, otherwise the browser's text editor, such as Firefox, would also use English for spell checking, and words in the language used would be incorrectly highlighted in red.

daysus:
## Summary
- Update the client to keep the document `<html lang>` attribute in sync with the active i18n language.
- This ensures browser spellcheck uses the selected interface/user locale instead of staying on English.

Fixes #1663

## Testing
- pnpm exec nx run client:build

### DIFF
--- a/apps/client/src/features/user/user-provider.tsx
+++ b/apps/client/src/features/user/user-provider.tsx
@@ -61,6 +61,10 @@ export function UserProvider({ children }: React.PropsWithChildren) {
   }, [data, isLoading]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


Philipinho:
Thanks for contributing.

PR #2169, opened 2026-05-08
fix: inline code text always rendered in pink/red

### LINKED ISSUE #1837: Code text always red
Hello, 

When writing simple code snippets, the code is always red and it’s not possible to change it to other colors.
On DocMost’s own website, syntax highlighting works as expected:

Image

But this is how it looks on our instance:

Image

We would love to see this behavior improved in future releases!

daysus:
Fixes #1837


Philipinho:
This was fixed. Thank you.

PR #2190, opened 2026-05-12
Better trash

notpeter:
I recently lost a bunch of time editing and searching for pages that were actually in the Trash. Docmost intentionally tries to not link to Trashed pages and only shows a read-only preview in Trash, but the url of the Trashed page and any inbound links still work.

This PR makes it clearer when a page you are interacting with is in the Trash and prevents you from editing trashed pages.

- /trash
  - Refactored banner into `trash-banner.tsx`
  - Refactored "Restore" modal into `use-restore-page-modal.tsx`
- Pages (when Deleted):
  - Add: `trash-banner.tsx`
  - Make read-only. Prevents editing. Hides "Move" and "Move to Trash" from menu.
  - Add "Restore" to menu (invokes `use-restore-page-modal`)
  - On restore/delete invalidate queries so trash banner immediately goes away.

I tried very hard to keep this as minimal as possible and not introduce new translation strings.

Before / After
docmost-trash-before-after

Potential future work:
- Add `"Failed to delete page"` to translation string (existing oversight)
- Change the display name so links on other Wiki pages to draw attention to linking to content pending deletion (e.g. "~Page Name~", "Page Name (Deleted)", "Page Name (Trash)", etc).

Let me know if you'd like me to do those as well.


Philipinho:
Thanks for contributing.

PR #2225, opened 2026-05-21
fix: prevent mobile keyboard on todo checkbox toggle

### LINKED ISSUE #2187: Toggling To-Do list checkboxes causes mobile keyboard to be opened
When I toggle a checkbox of a to-do list item on mobile/android, the virtual keyboard opens and my scroll position moves to the last line my cursor was at (no the same line that was toggled). This UX is particularly frustrating. This is especially painful since there is no edit-mode (Edit vs Read) where checkboxes can be toggled but text content is locked.

Expected/assumed behavior is that toggling a to-do list item checkbox should not enter text-editing mode and thereby opening the virtual/on-screen keyboard. It also should not cause my scroll position to move to the last location of my cursor, but thats likely a by-product of entering text-edit mode.

daysus:
## Problem

Tapping a task item checkbox on mobile devices triggers the virtual keyboard. The `editor.chain().focus()` call inside the `change` handler forces the editor to gain focus, which on mobile opens the on-screen keyboard and scrolls the viewport.

## Root cause

The stock `@tiptap/extension-list` TaskItem NodeView calls `editor.chain().focus()` on every checkbox `change` event (line 123 in the dist). On desktop this returns focus to the editor harmlessly. On mobile it pops the keyboard open because the tap already placed focus in the input element.

## Fix

- Created a custom TaskItem extension at `apps/client/src/features/editor/extensions/custom-task-item.ts`
- Identical schema to the stock one (same name, attributes, parse rules) for collaboration compatibility
- Only behavioral change: skips `.focus()` on touch-capable devices (`ontouchstart in window || navigator.maxTouchPoints > 0`)
- The checkbox toggle still works — it just doesn't re-focus the editor

## Files changed

- `apps/client/src/features/editor/extensions/custom-task-item.ts` — custom TaskItem with mobile-aware focus
- `apps/client/src/features/editor/extensions/custom-task-item.test.ts` — unit tests for extension config
- `apps/client/src/features/editor/extensions/extensions.ts` — imports custom TaskItem instead of stock

## Test evidence

Run: `cd apps/client && npx vitest run --reporter verbose`

```
 ✓ custom TaskItem extension > has the correct name
 ✓ custom TaskItem extension > has the checked attribute with default false
 ✓ custom TaskItem extension > parses data-checked attribute
 ✓ custom TaskItem extension > supports nested content
 ✓ custom TaskItem extension > defines content for nested mode
 ✓ custom TaskItem extension > defines content for non-nested mode
 ✓ custom TaskItem extension > keyboard shortcuts include Enter and Shift-Tab
```

## Compatibility

The custom extension produces the same ProseMirror schema as the stock TaskItem:
- Same node name: `taskItem`
- Same attributes: `checked`
- Same content expression: `paragraph+` / `paragraph block*`
- Same parse/render HTML

This ensures Yjs collaboration between clients using the custom extension and the server using the stock one remains fully compatible.

Fixes #2187


Philipinho:
The extension belongs in the `packages/editor-ext ` directory.
Also, we should [extend](https://tiptap.dev/docs/editor/extensions/custom-extensions/extend-existing) the existing extension instead of copying all its code.
We do these for a few other extensions.

PR #2259, opened 2026-05-31
fix: pdf table header alignment issue

### LINKED ISSUE #2248: [BUG] Table headers shifted down in PDF export
This behavior can be seen in all browsers I've test (Chrome, Brave) and is shown even when previewing the PDF before saving.

Happened on all pages I've tested.

Image

Mayank-2-16:
## Summary

Fixes #2248.

This PR fixes table header alignment in PDF/print output. Table headers could appear shifted down when exporting or previewing a page as PDF.

The issue was caused by PDF rendering reusing the normal read-only editor extension setup, which includes interactive table behavior intended for screen viewing. In particular, table header pinning and readonly table sorting can add sticky/relative positioning, transforms, chevrons, and extra header padding. Those behaviors are useful in the browser, but they should not affect static PDF output.

This PR adds a print-specific render mode for `ReadonlyPageEditor` and uses it from the dedicated PDF render page. In that mode, the editor skips the interactive table header extensions so tables render as static document content for PDF generation.

## Changes

- Added an optional `printMode` prop to `ReadonlyPageEditor`.
- When `printMode` is enabled, `ReadonlyPageEditor` excludes:
  - `tableHeaderPin`
  - `tableReadonlySort`
- Updated the PDF render page to pass `printMode` when rendering page content for PDF export.
- Added print CSS safeguards to:
  - reset pinned table header positioning during print
  - remove table header transforms during print
  - hide readonly sort chevrons in print output
  - restore normal header padding when sort chevrons are hidden

## Why This Approach

PDF output should be static and deterministic. Interactive table features such as sticky headers and sort controls are screen-only behavior, and they can interfere with browser print layout.

Instead of changing table behavior globally, this PR scopes the behavior change to the PDF render path. Normal page editing and normal read-only page viewing are preserved.

This keeps the fix narrow:

- editing behavior is unchanged
- normal readonly table sorting is unchanged
- normal table header pinning is unchanged
- PDF/print output no longer inherits interactive table UI behavior

The print CSS fallback is included to make browser print and Save as PDF more robust in case any table header pin/sort styles are already present when print media styles are applied.

## Manual Testing

Tested manually with a page containing a table with a real header row and body rows.

Verified:

- Table headers align correctly in PDF/print preview.
- Table headers are no longer shifted down in PDF output.
- Normal page viewing remains unaffected.
- Normal editor behavior remains unaffected.

Also ran:

```bash
pnpm --filter client build
```

The client build passed successfully.

##  Notes
This PR intentionally does not modify the server-side HTML/Markdown export service because the reported issue is in the browser PDF/print rendering path, not in HTML or Markdown export serialization.

### DIFF
--- a/apps/client/src/ee/pdf-export/pdf-render-page.tsx
+++ b/apps/client/src/ee/pdf-export/pdf-render-page.tsx
@@ -58,6 +58,7 @@ export default function PdfRenderPage() {
         title={data.title}
         content={data.content}
         pageId={data.pageId}
+        printMode
       />
     </Container>
   );
--- a/apps/client/src/features/editor/readonly-page-editor.tsx
+++ b/apps/client/src/features/editor/readonly-page-editor.tsx
@@ -15,6 +15,7 @@ interface PageEditorProps {
   title: string;
   content: any;
   pageId?: string;
+  printMode?: boolean;
   /**
    * When rendering inside a public share, pass the share's id (or key). Lookups
    * for transclusion content then resolve against the share graph instead of
@@ -28,6 +29,7 @@ export default function ReadonlyPageEditor({
   title,
   content,
   pageId,
+  printMode = false,
   shareId,
 }: PageEditorProps) {
   const [, setReadOnlyEditor] = useAtom(readOnlyEditorAtom);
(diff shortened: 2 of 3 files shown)


Philipinho:
Thanks for your contribution.

Issue #1945, opened 2026-02-14
Environment variable to configure default locale


michidk:
It would be great if there were an environment variable (or some other means of configuring) the default locale, so that not every user has to change it in their settings after logging infor  the first time.

PR #2266, opened 2026-06-05
feat: configurable default user locale via DEFAULT_LOCALE env

Sebastien-Gineste:
Based on https://github.com/docmost/docmost/issues/1945.

Adds a DEFAULT_LOCALE env variable to set the locale for newly created users. The locale was hardcoded to en-US in UserRepo.insertUser(), forcing every new user (signup, setup, invitation) into English regardless of the instance's audience.


virtualeye-dev-ops:
Please prioritise this if you can and enable non en-US english variants.
Thank you.

Issue #2254, opened 2026-05-28
i18n - labels don´t support unicode


thomasjaysmith:
<img width="397" height="193" alt="Image" src="https://github.com/user-attachments/assets/be45374e-b183-4508-a9a0-cf6202e95485" />

<img width="263" height="111" alt="Image" src="https://github.com/user-attachments/assets/53283a18-9042-4981-9fad-d2596ceefa09" />

Labels currently do not support non-ASCII characters.

For example, German umlauts such as `ä`, `ö`, `ü` are rejected as invalid label names. This likely also affects other languages, for example French characters like `é`, `è`, `à`, `ç`.

Expected behavior:
Labels should support Unicode characters so users can create localized labels in their own language.

Examples that should be valid:
- Schütz
- äääää
- Déjà vu
- Français
- Niño

Current behavior:
These labels are rejected with “Invalid label name”.

docmost 0.90.1

swaraj017:
Hey, I can take this one to solve,
I’ll look into the label validation logic and update it to support Unicode characters properly.

PR #2291, opened 2026-06-19
fix: the bug of the tags cannot be unicode

tanjingdong:
Bug Fix: Unicode Character Support for Labels — Extended the label name regex validation from ASCII-only to Unicode-compatible

Label Unicode Support
Issue: The name regex /^[a-z0-9_-][a-z0-9_~-]*$/ only permitted ASCII characters, causing labels in Chinese and other languages to fail with "invalid label name"
Fix: Updated to /^[\p{L}\p{N}_-][\p{L}\p{N}_~-]*$/u. \p{L} matches all Unicode letters (covering CJK, Arabic, etc.), \p{N} matches all Unicode digits, and the u flag enables Unicode mode
Affected Files: label.dto.ts (server-side validation) and label-picker.tsx (client-side validation)
Compatibility: Fully backward-compatible for ASCII labels; normalizeLabelName logic remains unchanged; database varchar columns natively support UTF-8 with no migration required

### DIFF
--- a/apps/client/src/features/label/components/label-picker.tsx
+++ b/apps/client/src/features/label/components/label-picker.tsx
@@ -16,7 +16,7 @@ type LabelPickerProps = {
   onClose: () => void;
 };
 
-const NAME_PATTERN = /^[a-z0-9_-][a-z0-9_~-]*$/;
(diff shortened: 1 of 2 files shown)


thomasjaysmith:
Closes  #2254

PR #2292, opened 2026-06-19
fix: subpage not loading issue

### LINKED ISSUE #2273: [Bug] Subpage list not updating on adding a new subpage
Steps to reproduce:
1. Create a page
2. In the page press / , choose  "Subpages (Child pages) - it's gonna show "No subpages"
3. Click [+] on the page to create a subpage, name it
4. Click on the page from step 1.
5. Check the subpage list - **it should update to show the newly created page but it doesn't**

Notes:
Pressing F5 updates the list.
Clicking around within the app, to other pages and settings etc., doesn't update the list.
Closing the tab and reopening the browser tab updates the list.
Amount of subpages during creation of the list doesn't affect the behavior - starting empty or with multiple subpages doesn't change the behavior.
The list updates on deleted pages, but it only cleans up deleted page, e.g. adding 2 new pages and deleting another doesn't make the other 2 show up, just removes the deleted page.

Mayank-2-16:
## Summary

Fixes: #2273 

The issue was caused by a React Query cache key mismatch. The Subpages block was reading child pages using a pageId-only sidebar query key, while page creation and related sidebar cache writers mostly updated the space-aware sidebar query key.

As a result, a newly created child page could exist in one sidebar cache entry while the Subpages block kept rendering stale data from another. Refreshing the browser rebuilt the cache and showed the correct result, which confirmed this was a frontend cache synchronization issue rather than a backend data issue.

## Changes

- Passed `spaceId` from the loaded page into the full editor flow.
- Stored both `pageId` and `spaceId` on the TipTap editor instance.
- Also stored both values on the temporary static editor render so the first pre-sync render uses the same page identity data.
- Updated `SubpagesView` to query sidebar child pages with `{ pageId, spaceId }` when `spaceId` is available.
- Preserved the existing shared-page Subpages behavior.
- Kept a pageId-only fallback for editor contexts where `spaceId` is unavailable.
- Added sidebar cache helpers so create, update, move, duplicate, and restore paths update matching sidebar child caches consistently.
- Added id-based deduplication when appending pages into sidebar cache data.
- Added regression tests for both sidebar query key shapes:
  - `["sidebar-pages", { pageId }]`
  - `["sidebar-pages", { pageId, spaceId }]`

## Why This Approach

The backend already returns the correct child pages. The stale UI happened because the Subpages block and mutation cache writers were not consistently using the same React Query cache entry.

Instead of changing backend APIs or forcing broader refetch behavior, this PR keeps the existing query strategy and updates the correct local sidebar caches when page mutations happen.

This keeps the fix narrow:

- no backend changes
- no global React Query default changes
- no CSS/layout changes
- shared-page rendering remains unchanged
- normal page tree behavior remains unchanged
- existing pageId-only sidebar cache entries are still handled defensively

## Testing

Ran:

```bash
pnpm.cmd --filter client test -- page-query.test.ts
pnpm.cmd --filter client build
```

Both passed.

## Notes

This PR intentionally does not modify the server-side page or sidebar APIs because the issue is caused by frontend React Query cache synchronization, not incorrect backend data.

It also avoids changing global React Query settings such as `refetchOnMount` or `staleTime`, since those settings affect the whole client and would be broader than needed for this bug.

The cache update helpers still handle pageId-only sidebar query keys defensively, even though the Subpages block now uses the space-aware key when available. This prevents stale sidebar cache entries from surviving if an older or fallback query shape is already present in the client cache.

### DIFF
--- a/apps/client/src/features/editor/components/subpages/subpages-view.tsx
+++ b/apps/client/src/features/editor/components/subpages/subpages-view.tsx
@@ -21,13 +21,20 @@ export default function SubpagesView(props: NodeViewProps) {
 
   //@ts-ignore
   const currentPageId = editor.storage.pageId;
+  //@ts-ignore
+  const currentSpaceId = editor.storage.spaceId;
 
   // Get subpages from shared tree if we're in a shared context
   const sharedSubpages = useSharedPageSubpages(currentPageId);
 
-  const { data, isLoading, error } = useGetSidebarPagesQuery(
-    shareId ? null : { pageId: currentPageId },
-  );
+  const sidebarParams = useMemo(() => {
+    if (shareId || !currentPageId) return null;
+    return currentSpaceId
+      ? { pageId: currentPageId, spaceId: currentSpaceId }
+      : { pageId: currentPageId };
+  }, [currentPageId, currentSpaceId, shareId]);
+
+  const { data, isLoading, error } = useGetSidebarPagesQuery(sidebarParams);
 
   const subpages = useMemo(() => {
     // If we're in a shared context, use the shared subpages
--- a/apps/client/src/features/editor/full-editor.tsx
+++ b/apps/client/src/features/editor/full-editor.tsx
@@ -48,6 +48,7 @@ export interface FullEditorProps {
   title: string;
   content: string;
   spaceSlug: string;
+  spaceId: string;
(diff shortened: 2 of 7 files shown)


Philipinho:
Fixed in v0.95.0.

PR #2301, opened 2026-06-20
Fix/issue 2276

### LINKED ISSUE #2276: [BUG] Cannot click on the checkbox icon
Hello,

I am creating this issue to report a bug where we cannot click on the checkbox icon to mark a task as completed. We are using Docmost under Docker on version 0.80.2
This happens on different web browse (Mozilla, Edge and Google Chrome). 

For some of our users it works but for most of them it doesn't. No error is occuring when we click on it and nothing is appearing in the "Network" tab on the inspector of the web page.

Image

I also confirm that nothing is logged on the Docker container when we click on the checkbox (no error or nothing).

I have checked the changelog of the latest release (because we dont have the latest one), though it doesn't seems like this is fixed in the version 0.90.1

Thank your for your help!

Abhigyan-git09:
### Description
Fixes #2276 

**Root Cause:** Tiptap renders task checkboxes inside a `<label>`. In some browser engines, applying `user-select: none` to this wrapper causes the browser or ProseMirror to swallow pointer events, preventing the click from reaching the input.

**Solution:** Updated `task-list.css` to add `pointer-events: none` to the wrapper label and `pointer-events: auto` to the checkbox. This allows clicks to bypass the wrapper and directly toggle the checkbox natively.

### DIFF
--- a/apps/client/src/features/editor/extensions/extensions.ts
+++ b/apps/client/src/features/editor/extensions/extensions.ts
@@ -214,6 +214,7 @@ export const mainExtensions = [
   TaskList,
   TaskItem.configure({
     nested: true,
+    onReadOnly: true,
   }),
   LinkExtension.configure({
     openOnClick: false,
--- a/apps/client/src/features/editor/styles/task-list.css
+++ b/apps/client/src/features/editor/styles/task-list.css
@@ -14,6 +14,7 @@ ul[data-type="taskList"] {
             flex: 0 0 auto;
             margin-right: 0.5rem;
             user-select: none;
+            pointer-events: none;
         }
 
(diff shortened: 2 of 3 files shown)


jfcmontmorency:
The fix only works in the editor. The problem persists on the shared page.

Abhigyan-git09:
Thanks for catching that! The shared pages use a read-only editor, which causes Tiptap to add a disabled attribute to the checkboxes, ignoring our CSS fix.

I've just pushed a commit setting onReadOnly: true in the TaskItem config to prevent it from disabling the inputs. Checkboxes are now clickable on shared pages as well. PTAL!

PR #2362, opened 2026-07-23
fix(sso-auth): Entra ID login fails silently when email claim is missing

Kaito-Vu:
## Summary
- Fall back to the `preferred_username` claim when Azure AD's ID token doesn't include `email` (only emitted when the optional claim is explicitly configured on the app registration) — previously this caused the OIDC callback to reject a successful token exchange with no visible error.
- Log the actual error/stack in the OIDC callback's catch block instead of silently redirecting to `/login?error=sso_failed`, so future failures are diagnosable from server logs.

## Why
A user reported a 500 during Entra ID login (fixed separately by correcting a tenant/client ID mix-up), then a blank white page after login succeeded at the token level. Tracing it against a known-working reference implementation showed our callback required the `email` claim unconditionally and swallowed all callback errors — a combination that fails invisibly for Entra ID tenants that don't emit `email` by default.

## Test plan
- [ ] `pnpm --filter ./apps/server exec tsc --noEmit` passes (done locally)
- [ ] Deploy and retry Entra ID login end-to-end; confirm successful redirect to `/` with a valid session
- [ ] If it still fails, confirm the new log line in `docker logs docmost-1` identifies the cause


Kaito-Vu:
Opened by mistake against the wrong repo (should have targeted a fork, not upstream). Closing.

PR #2398, opened 2026-08-15
feat: enhanced search filters

salihudickson:
closes: https://github.com/docmost/issue-tracker/issues/3

Screenshot 2026-08-15 at 22 51 12


Philipinho:
Thinking more about this, we have to handle such filters support for the Typesense integration as well.
That may mean indexing labels and users in Typesense too.

salihudickson:
yeah. Currently I implemented a bit of a work-around, but indexing the labels will definitely make the searching quicker. CreatorId for pages is already indexed, so no problems there.

Philipinho:
Thanks for your work.

Let's work on a follow up PR to further extend the labels visibility.

We can extend this by introducing new endpoints for label tag cloud, and a UI for it.
The space labels would be exposed at `/s/growth/labels` to display all the labels in that space as a tag cloud.

It should be accessed from the space menu.
image

At the workspace level, we do same with a new route at /labels. This route diplays all labels across all spaces the user has access to.
Ideally, follows the existing permissions we used for labels search.

PR #2404, opened 2026-08-18
fix(server): return 400 instead of 500 from shares/page-info when pageId is missing

### LINKED ISSUE #2396: shares/page-info returns 500 instead of 4xx when called with only shareId
### Describe the bug

`POST /api/shares/page-info` returns **500 Internal Server Error** when called with only `shareId` and no `pageId`.

The endpoint's own validation implies either field is acceptable:

```ts
// apps/server/src/core/share/share.controller.ts
if (!dto.pageId && !dto.shareId) {
  throw new BadRequestException();
}
```

and `ShareInfoDto` marks `shareId` as `@IsOptional()`. So a request carrying only `shareId` passes validation — then crashes.

### Root cause

`getSharedPage()` uses `dto.pageId` immediately and never reads `dto.shareId`:

```ts
// apps/server/src/core/share/share.service.ts
async getSharedPage(dto: ShareInfoDto, workspaceId: string) {
  const share = await this.getShareForPage(dto.pageId, workspaceId);   // <-- undefined
```

`undefined` reaches the postgres.js driver, which rejects it:

```
UNDEFINED_VALUE: Undefined values are not allowed
    at handleValue (postgres/cjs/src/types.js:83:20)
    at build (postgres/cjs/src/connection.js:230:25)
    at Object.execute (postgres/cjs/src/connection.js:172:7)
```

Because the throw happens in the driver rather than in a guard, Nest maps it to a 500 instead of a 4xx.

### To Reproduce

```bash
# 500 Internal Server Error
curl -X POST https://<host>/api/shares/page-info \
  -H 'content-type: application/json' \
  -d '{"shareId":"<share-key>"}'

# 200 OK - same share, pageId added
curl -X POST https://<host>/api/shares/page-info \
  -H 'content-type: application/json' \
  -d '{"shareId":"<share-key>","pageId":"<page-slugId>"}'
```

### Expected behavior

Either:

1. **Treat it as a client error** - if `pageId` is genuinely required, validate it up front and return 400 with a clear message. The guard becomes `if (!dto.pageId) throw new BadRequestException('pageId is required')`, and `shareId` should probably lose `@IsOptional()` ambiguity in the DTO.
2. **Honour `shareId` alone** - resolve the share by key and use its `pageId`, which is what the current validation appears to promise. This is friendlier for anyone hitting the API directly, since a share link exposes the key but not the page slugId.

I have not sent a PR because the choice between these is a product decision. Happy to implement whichever you prefer.

### Environment

- Reproduced on self-hosted **v0.95.0** (`docmost/docmost:latest`)
- Code path verified unchanged on `main` @ `c093c18`

### Notes

Not security-sensitive: no data is leaked, and the request fails closed. It is a robustness/DX issue - an unauthenticated public endpoint that 500s on input its own validator accepts.

vjymisal0:
Fixes #2396

`POST /api/shares/page-info` returns a 500 when called with only `shareId` (no `pageId`), because `getSharedPage()` only ever resolves the share by `dto.pageId` and never reads `dto.shareId`. The undefined `pageId` reaches the postgres.js driver and throws `UNDEFINED_VALUE`, which Nest maps to a 500 instead of a 4xx.

This PR keeps the existing behavior (resolve by `pageId`) but validates for it explicitly, so a request missing `pageId` fails closed with a 400 instead of crashing with a 500.

Added `share.controller.spec.ts` covering:
- only `shareId` provided → `BadRequestException`, service never called
- `pageId` provided → resolves normally through `ShareService`/`LicenseCheckService`

Verified against the pre-fix code that the new test fails with the same `TypeError`/undefined-value crash reported in the issue, and passes after the fix.

### DIFF
--- a/apps/server/src/core/share/share.controller.spec.ts
+++ b/apps/server/src/core/share/share.controller.spec.ts
@@ -1,0 +1,68 @@
+import { BadRequestException } from '@nestjs/common';
+import { ShareController } from './share.controller';
+import { ShareService } from './share.service';
+import { LicenseCheckService } from '../../integrations/environment/license-check.service';
+import { Workspace } from '@docmost/db/types/entity.types';
+import { ShareInfoDto } from './dto/share.dto';
+
+describe('ShareController - getSharedPageInfo', () => {
+  const buildController = () => {
+    const shareService = {
+      getSharedPage: jest.fn(),
+      isSharingAllowed: jest.fn(),
+    } as unknown as jest.Mocked<ShareService>;
+
+    const licenseCheckService = {
+      resolveFeatures: jest.fn().mockReturnValue({}),
+    } as unknown as jest.Mocked<LicenseCheckService>;
+
+    const controller = new ShareController(
+      shareService,
+      {} as any, // shareRepo
+      {} as any, // pageRepo
+      {} as any, // pagePermissionRepo
+      {} as any, // pageAccessService
+      licenseCheckService,
+      {} as any, // auditService
+    );
+
+    return { controller, shareService, licenseCheckService };
+  };
+
+  it('throws BadRequestException without touching the service when only shareId is provided', async () => {
+    const { controller, shareService } = buildController();
+    const dto = { shareId: 'some-share-key' } as ShareInfoDto;
+    const workspace = { id: 'workspace-1' } as Workspace;
+
+    await expect(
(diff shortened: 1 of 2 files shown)


vjymisal0:
recheck

Philipinho:
Thanks for contributing. This has been fixed in another PR.

PR #2413, opened 2026-08-20
fix: rework filtering behaviour

### LINKED ISSUE #2407: Multi-select behaviour in filters (usability)
If I create a new filter based on a select-field and I chose either “Is any of“ or “Is none of“, I cannot select multiple values, but it seems that I have to add multiple filter lines of the same type (which needs a quite complicated interaction, because you have to save the filter in between every time).

### Proposed change

Allow multi-select within the same filter-rule (checkbox, CTRL-select, comma-separated …)

salihudickson:
closes: #2407


salihudickson:
@Philipinho I changed the filtering UX so users have to save changes before they apply. This slightly reduces unnecessary API calls and avoids UI flashes from the base component unmounting and remounting. 

Not sure if you agree with this or if you prefer the way it was before?

Philipinho:
LGTM!

PR #2439, opened 2026-08-27
#2439 - Fix PDF files not rendering when embedded

Akshayrs46:
The default-src 'self' directive blocked the browser's built-in PDF viewer from loading internal resources (like blob: URLs), resulting in an empty iframe when rendering PDF attachments. This commit introduces a permissive CSP (omitting default-src) specifically for PDF responses, while maintaining the strict CSP for all other file types.


Philipinho:
Which browser did you experience this on?

PR #2441, opened 2026-08-27
feat: include page comments in AI context (#2441)

Akshayrs46:
Pass pageId to EditorAiMenu and build a markdown context block containing unresolved, top-level page comments and their replies from the React Query cache. This allows the AI to reference and validate ongoing page discussions when generating responses.


Philipinho:
Thanks for your contribution but this is not what the issue was about. In addition, the issue was resolved.

PR #2506, opened 2026-09-23
fix: avoid leading hyphen in page slug when title transliterates to empty

### LINKED ISSUE #2397: Page URLs start with a stray hyphen when the title transliterates to an empty slug (CJK, Hebrew)
### Describe the bug

For page titles in scripts that `@sindresorhus/slugify` cannot transliterate, `buildPageSlug()` produces an empty title segment, so every generated URL carries a **leading hyphen**:

```
/s/general/p/-x7Kd91mQa
/share/<key>/p/-x7Kd91mQa
```

### Cause

```ts
// apps/client/src/features/page/page.utils.ts
const buildPageSlug = (pageSlugId: string, pageTitle?: string): string => {
  const titleSlug = slugify(pageTitle?.substring(0, 70) || "untitled", { ... });
  return `${titleSlug}-${pageSlugId}`;   // titleSlug can be ""
};
```

When `titleSlug` is `""` the template still emits the separator.

### Verified behaviour (`@sindresorhus/slugify@3.0.0`, the pinned version)

| Title | `titleSlug` | Resulting URL segment |
| --- | --- | --- |
| `Hello World` | `hello-world` | `hello-world-AbC123` |
| `Привет мир` | `privet-mir` | `privet-mir-AbC123` |
| `لماذا نعرّب تطبيقاتنا` | `lmadha-nerb-ttbyqatna` | `lmadha-nerb-ttbyqatna-AbC123` |
| `测试页面` | `""` | **`-AbC123`** |
| `שלום עולם` | `""` | **`-AbC123`** |

So Arabic and Cyrillic transliterate fine; CJK, Hebrew, Thai and similar scripts hit the empty case.

### Impact

Cosmetic only. `extractPageSlugId()` splits on `-` and takes the last segment, so these URLs resolve correctly - they just look malformed when shared, and the leading `-` is easy to lose in chat clients that auto-link.

### Suggested fix

```diff
-  return `${titleSlug}-${pageSlugId}`;
+  return titleSlug ? `${titleSlug}-${pageSlugId}` : pageSlugId;
```

A bare `pageSlugId` is already a valid input to `extractPageSlugId()` (it returns `slug` unchanged when there is no `-`), so no parsing changes are needed.

Happy to open a PR if you would like this - just let me know, since I would need to sign the CLA first.

### Environment

- `main` @ `c093c18`
- `@sindresorhus/slugify` `3.0.0` (root `package.json`)

f1n3d:
When `@sindresorhus/slugify` cannot transliterate a title (CJK, Hebrew, Thai, …) it returns `""`, so `buildPageSlug()` produced `/-x7Kd91mQa`-style URLs with a leading hyphen.

This changes all three slug builders to return the bare `slugId` when `titleSlug` is empty:
- `apps/client/src/features/page/page.utils.ts`
- `apps/server/src/integrations/export/export.service.ts`
- `apps/server/src/integrations/import/utils/import-formatter.ts`

`extractPageSlugId()` already returns the input unchanged when there is no `-` (and UUIDs directly), so old and new URLs resolve identically — cosmetic fix only.

Fixes #2397


salihudickson (commented):
need to add the same fix for the confluence-import.service module, other than that LGTM.

PR #2507, opened 2026-09-23
fix: serve share attachments for absolute urls and pasted files

### LINKED ISSUE #2502: Images return 401 on publicly shared pages, even when uploaded directly to that page
**Version:** v0.95.0, self-hosted. I checked the v0.96.0 release notes and did not see anything covering attachment authorisation on share links, but I have not tested against it.

### Summary

A publicly shared page serves its own HTML anonymously (200), but every attachment it references returns `401 Unauthorized`. Anyone opening the share link without an account sees the page with all images broken.

This looks like #1194, which was closed on the basis that it happens when *"the attachment was copied and pasted from another page"*. That does not apply here: every image in these pages was uploaded directly to the page that displays it, via `POST /api/attachments/upload-file` with that page's id. #1835 also reports share-link images failing (for images inside tables), so something in this area still appears broken.

### Steps to reproduce

1. Create a page.
2. Upload an image directly to that page (not copied from elsewhere).
3. Share the page publicly.
4. Open the share link in a private window.

### Expected

The image renders, as the page itself does.

### Actual

The page renders; the image is broken.

```
GET /share/<shareId>/p/<slug>              → 200
GET /api/files/<uuid>/<filename>           → 401 {"message":"Unauthorized","statusCode":401}
GET /api/files/<uuid>/<filename>           → 401   (with Referer set to the share page)
```

No public attachment route appears to exist. These all return the SPA shell rather than the file:

```
GET /files/<uuid>/<filename>                     → 200 text/html (1667 bytes)
GET /api/attachments/<uuid>/<filename>           → 200 text/html (1667 bytes)
GET /share/<shareId>/api/files/<uuid>/<filename> → 200 text/html (1667 bytes)
```

### Notes

The `/api/files/<uuid>/<filename>` URL form is Docmost's own: pages authored by hand in the editor produce the identical shape, and they break the same way on a share link.

Related: #1194 (closed), #1835 (open).

f1n3d:
Publicly shared pages returned `401` for images even when uploaded directly to the page.

**Root causes**
1. `updateAttachmentAttr()` only rewrote relative `/api/files` urls (`startsWith`). Older docs store absolute urls (`https://host/api/files/...`, cf. #1194) which were left as private links, so anonymous viewers requested the auth-guarded endpoint → `401`. Now matches via `includes` and preserves the host prefix.
2. Public tokens were always scoped to the host page, but the file endpoint enforces `attachment.pageId === token.pageId`. Files pasted from another page (or older docs, cf. #1194) are still owned by their original page → rejected. Tokens are now scoped to the attachment's actual owner page (same workspace, DB lookup), falling back to host/source page.

Also skips already-public urls to avoid double `/public/public` rewrites.

**Changed**
- `apps/server/src/core/share/share.util.ts`
- `apps/server/src/core/share/share.service.ts` (`prepareContentForShare`)

Fixes #2502


apps/server/src/core/share/share.util.ts
```diff
@@ -8,7 +8,8 @@ export function updateAttachmentAttr(
   const attrVal = node.attrs[attr];
   if (
     attrVal &&
-    (attrVal.startsWith('/files') || attrVal.startsWith('/api/files'))
+    !attrVal.includes('/files/public/') &&
+    (attrVal.includes('/api/files/') || attrVal.includes('/files/'))
```
salihudickson: this is a bit too expansive and can cause false positives. We can just strip the host and then run then use the `.startsWith` check

salihudickson: this is a bit too expansive and can cause false positives. We can just strip the host and then run then use the `.startsWith` check.

apps/server/src/core/share/share.service.ts
salihudickson: Good catch on this. But this is an unsafe fix, basically we'd be giving access to assets on a different page even if the page is private.

